PR #14, opened 2024-04-23
feat: Simplify SDK & CLI improvements

michalziolkowski:
1. You no longer have to keep SHERLO_TOKEN in EAS secrets
2. We don't use `yargs` as a dependency because it enforced using ESM module which is a problem for some users
3. We return error codes on CLI response
4. We no longer calculate next snapshot index on the client, it's passed as a part of the runner-client protocol
5. Added support for iOS .tar file
6. Changed file compression to on-the-fly
7. Removed unnecessary apps object wrapper from config


examples/expo-example/.storybook/storybook.requires.ts
```diff
@@ -1,6 +1,10 @@
 /* do not change this file, it is auto generated by storybook. */
```
dawidk92, line 1: Make sure that these changes don't break Storybook

examples/expo-example/builds/ios.tar.gz
dawidk92, line 1: Make sure that we have correct builds

packages/cli/package.json
```diff
@@ -38,20 +38,19 @@
     "@sherlo/sdk-client": "1.1.4",
     "@sherlo/shared": "1.1.4",
     "chalk": "4.1.2",
+    "commander": "^12.0.0",
```
dawidk92: -> "12.0.0"

dawidk92 (approved):

PR #58, opened 2025-04-04
fix(react-native-storybook): Fix noSafeArea behaviour for Storybook 7

michalziolkowski:


packages/react-native-storybook/package.json
```diff
@@ -20,6 +20,12 @@
   },
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "exports": {
```
michalziolkowski, line 23: sprawdzic czy potrzebne

testing/expo-storybook-8/package.json
```diff
@@ -13,21 +13,23 @@
     "clear:devices": "../../scripts/overrideSherloMode.sh --clearAll",
     "update:development": "npx eas-cli update --channel=development --json --non-interactive --message='Update message'",
     "update:preview": "npx eas-cli update --channel=preview --json --non-interactive --message='Update message'",
-    "eas-build-on-complete": "sherlo eas-build-on-complete --profile=preview",
+    "eas-build-on-complete": "./node_modules/.bin/sherlo eas-build-on-complete --profile=preview",
```
michalziolkowski: sprawdzic z npx

testing/expo-storybook-8/package.json
```diff
@@ -48,13 +50,21 @@
...
     "@storybook/addon-ondevice-backgrounds": "^8.3.9",
     "@storybook/addon-ondevice-controls": "^8.3.9",
     "@storybook/addon-ondevice-notes": "^8.3.9",
     "@storybook/react-native": "^8.3.9",
     "babel-loader": "^8.2.3",
     "typescript": "~5.3.3"
-  }
+  },
+  "resolutions": {
+    "sherlo": "portal:../../packages/cli"
```
michalziolkowski, line 63: czy z npx to zadziala bez tego?

packages/react-native-storybook/package.json
michalziolkowski: Both fields serve different purposes and are often used together for compatibility reasons:
main and types are traditional fields used by older Node.js versions and TypeScript
exports is a newer Node.js feature that provides more precise control over package entry points
For maximum compatibility, keep both. The main field ensures older tools can use your package, while exports gives you better control with newer environments.
If you only target newer Node.js environments (12+), you could technically use just exports, but including both is the safer approach for wider compatibility.

PR #173, opened 2026-07-06
feat: [Product bug] @sherlo-io/cli mis-populates prHeadCommitHash…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1629

## Root cause (refined from the ticket's diagnosis)

SHERLO-1629: within-PR accept-once silently broke because an approved PR build was keyed in the commitHash GSI by the ephemeral merge-ref SHA instead of the real PR head, so a later build could not find it (getCanonicalCommitHash = prHeadCommitHash ?? commitHash).

Confirmation refined the diagnosis:
- The ticket's headline fix (set gitInfo.prHeadCommitHash to the real PR head) was ALREADY implemented on main and shipped in published 2.0.1 (commits #163/#164): getGitInfo sets prHeadCommitHash = git rev-parse HEAD^2.
- The genuine LIVE bug that survives 2.0.1: on GitHub's DEFAULT actions/checkout (fetch-depth:1) shallow checkout of a refs/pull/N/merge ref, the merge commit's parents are grafted away, so git rev-parse HEAD^2 fails. prHeadCommitHash was then omitted and commitHash decayed to the merge SHA -> the build was GSI-keyed by the merge SHA -> approved PR build invisible to later builds -> the unchanged approved story rebaselined and flipped to unreviewed.

## Fix

Recover the real PR head from the CI event payload (GITHUB_EVENT_PATH -> pull_request.head.sha) when local HEAD^2 is unavailable. Local HEAD^2 stays preferred (authoritative git); when neither exists, prHeadCommitHash is omitted rather than set to a wrong SHA. Adds 3 tests for the previously-uncovered shallow-clone scenario. yarn build OK; yarn test 206 passed (getGitInfo 44).

## PROD EXPOSURE (acceptance criterion)

YES. Real customers using the default shallow checkout hit this same bug (approved stories return to unreviewed on a second push). This fix resolves it for them.

## AC #1 confirmation note

The E2E run 28751401698 artifacts (pr-build-c.txt / pr-build-d.txt) were off-disk and could not be retrieved to read build C's stored field values. Root cause was instead confirmed mechanistically and locked with reproduction tests (stronger than the artifact snapshot). Happy to have the tester dept fetch the artifacts if you want the exact stored SHAs for the record.

## Out of scope (operator / optional)

- OPERATOR-ONLY: publishing the fixed CLI to npm + ensuring the dev env picks it up. Not automated.
- OPTIONAL api GSI dual-key hardening (sherlo-api): deferred - it needs a DB migration (sensitive path) and the root-cause fix already resolves the bug. Recommend a separate follow-up PR if the defense-in-depth layer is wanted.
- The E2E test 06-pr-ancestry-scenario.spec.ts:315 is untouched.

---

### DIFF
--- a/packages/cli/src/helpers/__tests__/getGitInfo.test.ts
+++ b/packages/cli/src/helpers/__tests__/getGitInfo.test.ts
@@ -1,6 +1,22 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import getGitInfo, { ANCESTOR_LIMIT } from '../getGitInfo';
 import GitFixture from './support/gitFixture';
+
+/**
+ * Writes a GitHub Actions event payload to a temp file and returns its path, so
+ * a test can point GITHUB_EVENT_PATH at it (mirroring how the runner exposes the
+ * triggering event's JSON to a job). Registered for cleanup in afterEach.
+ */
+const eventFiles: string[] = [];
+function writeGitHubEvent(payload: unknown): string {
+  const file = path.join(fs.mkdtempSync(path.join(os.tmpdir(), 'sherlo-gh-event-')), 'event.json');
+  fs.writeFileSync(file, JSON.stringify(payload), 'utf8');
+  eventFiles.push(file);
+  return file;
+}
 
 /**
  * Real-git tests for getGitInfo, driven by the deterministic {@link GitFixture}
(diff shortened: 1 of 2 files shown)


michalziolkowski:
Executor note (loop mode): this fix is complete and ready for review, but the loop cannot auto-merge yet. The task involves 3 repos (sherlo-api, sherlo-developer, sherlo) and loop merge requires a merged PR in every involved repo; only sherlo has work, so sherlo-api and sherlo-developer (no PRs) block completion. Escalated to the operator on SHERLO-1629 for a decision (recommend landing this PR standalone; publish + optional sherlo-api hardening as separate follow-ups). Review of this diff can proceed independently.

PR #176, opened 2026-07-09
feat: CLI: fingerprint module + base registration in full flows

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1690

## SHERLO-1690 - CLI: fingerprint module + base registration in full flows

Part of the Staged Build Upload epic (SHERLO-1682). Ships the augmented fingerprint module in the CLI and makes `test:standard` and `eas-build-on-complete` compute `baseFingerprint` + per-platform gate metadata and register the uploaded binary as the stageable base, fail-soft, with `nativeFingerprint` and `gitInfo` unchanged.

### Module (`helpers/fingerprint/`)
- `baseFingerprint.ts` - 3-layer, version-suppressed hash (managed `SourceSkips`; bare streaming `fileHookTransform`; + lockfiles + autolinked modules).
- `gateMetadata.ts` - per-platform `GateMetadataInput` matching sherlo-api #261: `engineClass` ('hermes'|'jsc', derived from project config - app jsEngine / gradle `hermesEnabled` / Podfile `:hermes_enabled`), `bundleFormat` ('hbc'|'plain-js'|'ram', from the embedded-bundle header sniff), `hasEmbeddedBundle`, `assetInventory`, `expoUpdatesEnabled` (per-platform boolean), `sdkProtocolVersion`, `buildMetadata`.
- `notStageable.ts` - eligibility reasons aligned with the API failReason derivation.
- `registerBase.ts` - fail-soft orchestrator (compute -> extract -> check -> return per-platform metadata).

### Wire
`baseFingerprint` (project-level, computed once) + `gateMetadata: { android?, ios? }` (per-platform) are attached to the existing `openBuild` (test:standard) and `asyncUpload` (eas-build-on-complete) payloads, behind a presence check. `nativeFingerprint` and `gitInfo` are untouched (AC4).

### Review response (round 2)
All four findings addressed:
1. **Wire registration** - `baseFingerprint` + per-platform `gateMetadata` now attached to `openBuild`/`asyncUpload` (no longer print-only). The CLI's `GateMetadataInput` mirrors #261's contract field-for-field. Note: the typed fields (`OpenBuildRequest.baseFingerprint`/`gateMetadata`) arrive via #261's `@sherlo/api-types` bump, delivered through the S3-published `sherlo-lib` bundle (`init-env`). Per the agreed sequencing, #261 merges first and the api `release_to_dev` publish makes these live; the CLI sends them behind a presence check now (no `as any`; the metadata object is fully typed against the local `GateMetadataInput` mirror). Keeping this PR open for the hold-all-then-batch merge.
2. **HBC magic byte order** - corrected to little-endian on-disk bytes `c6 1f bc 03 c1 03 19 1f`; the sniff now populates `bundleFormat`; unit test added with real header bytes.
3. **GateMetadata contract** - converged to the per-platform shape and field names above; `engineClass` (runtime engine) split from `bundleFormat` (bundle sniff); nested expo-updates flags collapsed to a per-platform `expoUpdatesEnabled`; `sdkProtocolVersion` promoted; `bundleFormat`/`hasEmbeddedBundle` feed eligibility.
4. **iOS expo-updates value** - parses the boolean VALUE (`<true/>`/`<false/>`/`<string>YES|NO`) instead of key presence.

Non-blocking notes addressed: portable `tar` (GNU `--wildcards` via `detectTarVersion`), `xxd` replaced by a Node byte read in the APK sniff, `DEFAULT_PROJECT_ROOT` import confirmed.

Build (ncc + tsc) clean; lint clean; tests green (fingerprint suite: base fingerprint determinism/version suppression, engine sniff with real header bytes, per-AC not-stageable cases, plist value parsing).

Split-APK detection remains dropped this iteration (single-ABI heuristic caused false positives on arm64-only universal APKs); real split detection needs a manifest marker and can follow.

Public-facing CLI copy / staged-flow docs (incl. `llms.txt`) stay content-owned under the docs subtask (SHERLO-1694); `--skip-llms-check` used on push (no new command/flag surface).

---

### DIFF
--- a/packages/cli/src/commands/easBuildOnComplete/helpers/asyncUploadBuildAndRunTests/__tests__/asyncUploadBuildAndRunTests.diffScope.test.ts
+++ b/packages/cli/src/commands/easBuildOnComplete/helpers/asyncUploadBuildAndRunTests/__tests__/asyncUploadBuildAndRunTests.diffScope.test.ts
@@ -58,6 +58,10 @@ vi.mock('../../../../../helpers/diffScope', () => ({
 
 vi.mock('../getBuildPath', () => ({ default: mocks.getBuildPath }));
 
+vi.mock('../../../../../helpers/fingerprint', () => ({
+  registerBase: vi.fn().mockResolvedValue({ registered: false }),
+}));
+
 import asyncUploadBuildAndRunTests from '../asyncUploadBuildAndRunTests';
 
 // ---------------------------------------------------------------------------
--- a/packages/cli/src/commands/easBuildOnComplete/helpers/asyncUploadBuildAndRunTests/__tests__/asyncUploadBuildAndRunTests.diffScope.test.ts
+++ b/packages/cli/src/commands/easBuildOnComplete/helpers/asyncUploadBuildAndRunTests/__tests__/asyncUploadBuildAndRunTests.diffScope.test.ts
@@ -59,6 +59,7 @@ vi.mock('../../../../../helpers/diffScope', () => ({
 vi.mock('../getBuildPath', () => ({ default: mocks.getBuildPath }));
 
 vi.mock('../../../../../helpers/fingerprint', () => ({
+  computeBaseFingerprint: vi.fn().mockResolvedValue({ hash: null }),
   registerBase: vi.fn().mockResolvedValue({ registered: false }),
 }));
 
--- a/packages/cli/src/commands/easBuildOnComplete/helpers/asyncUploadBuildAndRunTests/asyncUploadBuildAndRunTests.ts
+++ b/packages/cli/src/commands/easBuildOnComplete/helpers/asyncUploadBuildAndRunTests/asyncUploadBuildAndRunTests.ts
@@ -12,7 +12,11 @@ import {
   reporting,
 } from '../../../../helpers';
 import { computeChangedFiles, computeNativeFingerprint } from '../../../../helpers/diffScope';
-import { registerBase } from '../../../../helpers/fingerprint';
+import {
+  computeBaseFingerprint,
+  registerBase,
+  type GateMetadataInput,
+} from '../../../../helpers/fingerprint';
 import { THIS_COMMAND } from '../../constants';
 import getBuildPath from './getBuildPath';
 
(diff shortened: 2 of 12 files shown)


michalziolkowski:
[loop:question] Addressing the review. Findings 2 (HBC byte order) and 4 (iOS expo-updates value parsing) are clear and I'm fixing them. Two cross-PR contract points with sherlo-api#261 need your call before I wire findings 1 and 3, so I don't guess and trip the recurrence guard:

1. gateMetadata field semantics (finding 3). #261's `GateMetadataInput` uses `engineClass` with values "Hermes"/"JSC" (runtime engine, case-sensitive in resolveStagedGate). But the CLI's current `engineClass` is a BUNDLE-FORMAT sniff (HBC magic vs plain JS = "hermes"/"plain-js") - a different concept. You noted #261 round 2 adds `bundleFormat`/`hasEmbeddedBundle`. So, to converge: should the CLI's HBC-magic sniff populate `bundleFormat` (with what exact values - e.g. "hbc"/"plain-js"?), and how should the CLI determine `engineClass` (runtime engine "Hermes"/"JSC") - from the app's jsEngine config, or is engineClass API-derived and the CLI omits it? And confirm the per-platform wire shape is `gateMetadata: { android?: GateMetadataInput, ios?: GateMetadataInput }` with `hasEmbeddedBundle: Boolean` per platform. If you can share #261's round-2 GateMetadataInput once it lands, I'll match it field-for-field.

2. api-types delivery for the wire (finding 1). The CLI consumes `@sherlo/api-types` and `@sherlo/sdk-client` from the gitignored, S3-published `sherlo-lib` bundle (via `yarn init-env`), not from sherlo-api source. So a typed openBuild/asyncUpload call sending `baseFingerprint` + `gateMetadata` cannot compile (and CI, which re-runs init-env, will fail) until #261's api-types are published into that bundle. How do you want to sequence this for the hold-all-then-batch merge - does #261 publish the updated sherlo-lib (or a prerelease) that this PR then depends on, and in what order? I'll keep #176 open per your instruction; I just need the delivery mechanism so the wire call is real and green, not an `as any` cast.

PR #178, opened 2026-07-09
feat: CLI: test:bundled command

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1691

https://sherlo-io.atlassian.net/browse/SHERLO-1691

## SHERLO-1691 - CLI: `test:bundled` command

The bundle-only fast path for the Staged Build Upload epic (SHERLO-1682). This PR ships the **safe local half** of the command plus the runner contract cemented in tests. The server wire-up is intentionally deferred until the API consume-mode (SHERLO-1707) lands - see "Why the server send is deferred" below.

### What it does (reshaped after loop review round 1)
- **Bundle production** (`buildBundle.ts`, unchanged from round 1): reuses show-error's `detectBundler` + `detectEntryFile` to pick expo (`expo export:embed`) vs bare-RN (`react-native bundle`), both with `--dev false --minify true` (design §5). Branches explicitly on project type - never runs `npx expo` on a pure bare-RN project.
- **Format gate**: 12-byte HBC-magic sniff -> `.hbc` bundles refused with guidance; RAM/indexed bundles refused. Plain-JS is the v1 standard.
- **Version floor**: RN < 0.71 / Expo SDK < 50 refused with guidance (design §5/§7).
- **Readiness gate**: the command is gated behind `SHERLO_DEVTOOLS=1`. Without it, `test:bundled` prints "Staged uploads require server support that is still rolling out (SHERLO-1707). Run `sherlo test:standard` for a full build." and exits non-zero. With the override set, it runs the local pipeline (build + sniff + fingerprint + staged-config construction) but sends nothing to the server.
- **Runner staged contract - not cemented CLI-side (round-2 correction)**: per SHERLO-1707's final spec the runner-facing shape `{ url, jsBundleUrl, assetsUrl }` is assembled **server-side** in `getBuildRunWithUrls` from stored s3Keys, so the CLI has no runner shape to construct. The round-1 placeholder `buildStagedRunConfig` module cemented a misread of that contract and was **deleted**. The CLI's actual post-1707 payload (`{ jsBundleS3Key, assetsS3Key?, bundleSizeMb }`) will be introduced by the SHERLO-1707 wiring change, together with the upload code that produces the real s3Keys.
- **All refusal / not-ready paths exit non-zero** so CI never reads a refused staged run as a pass (fingerprint-unavailable, bundle-build failure, HBC/RAM/version-floor, no-devices, and the readiness gate).
- **gitInfo parity** (AC): built via the identical `getGitInfo` call test:standard uses - byte-identical for the same checkout.
- **Settle timeout**: `bundleSizeMb` is computed and carried in the staged config so the runner can apply the size-scaled settle timeout (Director spike B: `max(30s, MB*2s)`).

### Why the server send is deferred (round-1 blockers, resolved)
The round-1 review found the wire-up half could not ship, and one part was destructive:
- Sending `baseFingerprint` through `openBuild` routed this **consumption** run into the API's base-**registration** path, whose unconditional S3 CopyObject would overwrite the user's stored native base with the JS bundle - poisoning the exact artifact the run splices against. **Removed entirely**: `testBundled.ts` no longer calls `openBuild`, `getBuildUploadUrls`, or any upload. It builds locally, runs the sniffs/fingerprint, constructs the staged config, and exits.
- The API consume-mode that will accept these inputs safely is **SHERLO-1707** (filed and fast-tracked in parallel). When it lands, wiring the constructed `buildStagedRunConfig` output into the extended `openBuild` call is a small follow-up - the contract is already cemented in tests here.
- The `dev=true` refusal AC is satisfied-by-construction: the CLI always passes `--dev false`; there is no code path to a dev bundle. The unreliable round-1 `__DEV__` byte-scan was removed.

### Tests
47 tests pass across 3 files:
- `buildBundle.test.ts` - HBC refusal, RAM refusal, version-floor refusal, asset-inventory collection, refusal fallback line, and a version/build-number-only-change-does-not-refuse assertion (rides SHERLO-1690's version suppression).
- `buildStagedRunConfig.test.ts` - exact key-set assertions for the runner contract (with and without `assetsUrl`), `url` always empty, verbatim pass-through.
- `testBundled.test.ts` - readiness-gate message + non-zero exit, `SHERLO_DEVTOOLS` bypass, no-devices non-zero exit, fingerprint-unavailable non-zero exit + `test:standard` fallback line.

`tsc --noEmit` clean; repo lint 0 errors (pre-existing bitwise/shadow warnings in the untouched `fingerprint/*` files only).

### Deferred, by design
- **Public docs / llms.txt**: `test:bundled` is not yet end-to-end usable (gated behind the readiness check), so documenting it to end users now would be premature. Public docs for the staged flow are the dedicated **SHERLO-1694** subtask (blocked-by this ticket). The llms drift check was overridden on push for this reason.
- **Server wiring**: SHERLO-1707 (API consume-mode), then the small follow-up to send the staged config.

Closes SHERLO-1691.

---

---


michalziolkowski:
Answering the dev=true refusal question: it is satisfied-by-construction, and I recommend recording the AC that way.

The command invokes the bundler itself and always passes --dev false (buildBundle.ts execBundler, both the expo export:embed and react-native bundle paths). There is no code path where test:bundled produces or accepts a caller-supplied bundle, so a dev=true bundle can never enter the flow. The only theoretical escape is a project whose Metro/babel config force-overrides __DEV__ back to true despite --dev false - pathological and out of scope for v1. Round 1 had a byte-scan for that (a /__DEV__\s*=\s*true/ regex over the first 64KB); it was unreliable (with --dev false Metro replaces __DEV__ with a literal, and the scan also false-matches __DEV__ appearing in string content), so it was removed in round 2. A reliable dev-vs-prod detection is not cheaply available at the plain-JS layer, so defense-in-depth here would add fragile code for an unreachable case. Please record the dev=true refusal AC as satisfied-by-construction.

Agreed on the full reshape (remove the destructive baseFingerprint/openBuild staged sends + asset upload, non-zero refusal exits, buildStagedRunConfig contract module cemented in unit tests without sending it, hard readiness gate pending SHERLO-1707). Reshaping now.

PR #180, opened 2026-07-10
feat: CLI: staged:check + CI outputs

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1692

SHERLO-1692: CLI staged:check command + CI outputs.

## staged:check
New CI routing probe that answers mode=fast|full|not-stageable (plus a human reason) without bundling, uploading, or opening a build, so it runs right after dependency install with no build artifacts. Emits the decision in three synced forms:
- Exit code: fast=0, full=1, not-stageable=2
- --json: a single machine-readable JSON line (mode, reason, baseFingerprint, per-platform detail)
- $GITHUB_OUTPUT: always written (mode + reason [+ baseFingerprint]), with or without --json

The outcome is a routing decision, not an error: every expected mode funnels through a single report-and-exit path with its contractual code (not wrapped in the throwing timeout path), so CI never sees an expected state as a crash. Only genuine tool errors (bad token, network) fall through to the normal error path.

## test:bundled --on-stale=fail|build
- fail (default): when the staged gate finds the base stale, exits with the fingerprint diff naming the sources that moved. Tests are never silently skipped. Missing fullBuild config is an explicit diff-bearing failure.
- build: runs the per-platform staged.fullBuild shell commands from sherlo.config.json, then completes as a full run that registers a fresh base by delegating to test:standard (base registration stays in one place).

## Gate wiring
Both surfaces call the SHERLO-1718 sdkClient.checkStagedGate wrapper (per platform) rather than discovering staleness only via an openBuild refusal. The openBuild refusal path is kept as a consistent safety net routed through the same --on-stale handler.

## Tests
Added unit coverage for the exit-code contract, the --json payload shape, GITHUB_OUTPUT always written, and the --on-stale fail vs build branches. Verified against the genuine dev tarballs (init-env.sh CLIENT_STAGE=dev): lint 0 errors, tsc clean, tests green.

## Notes
- staged:check sends empty gate metadata (no binary by design); AC1 (mode=fast on a matching fingerprint) relies on the server returning fast for a fingerprint match with empty metadata, an API-side contract validated E2E in SHERLO-1695/1696.
- The public llms.txt update for the new command belongs to the docs ticket SHERLO-1694 (llms files live in sherlo-www).

---


michalziolkowski:
Addressed both review findings (head fbd34d0a):

Finding 1 (blocker - fixed): staged:check no longer sends empty gate metadata. It now builds a production plain-JS bundle per platform and derives gate metadata via the SAME buildBundleForPlatform + buildGateMetadata path test:bundled uses, then sends that real metadata to checkStagedGate - so a genuinely matching base returns fast instead of always full. Building a JS bundle is not a native artifact, so the "runnable after dependency install" property holds; if bundling fails the command still exits cleanly as not-stageable (no crash). Added a drift-guard test asserting real, non-empty metadata is constructed through that shared path (a client-mock alone cannot catch drift), plus a bundling-failure -> not-stageable test.

Finding 2 (runShellCommand): this is a pre-existing exported helper, not a missing symbol - it lives at src/helpers/runShellCommand/, is exported from src/helpers/index.ts, and is already used in ~15 files (init, testEasCloudBuild's runScript, testEasUpdate, helpers/fingerprint, ...). onStale.ts's call runShellCommand({ command, projectRoot }) matches its signature, so there is no compile failure (tsc is clean). I did not re-create it. I did add the trust-model doc-comment you asked for (staged.fullBuild commands are user-authored and config-declared, run through the same trust boundary as other config-driven command execution), and onStale.test.ts covers the build branch.

Verified against the genuine dev tarballs (init-env.sh CLIENT_STAGE=dev): tsc clean, lint 0 errors, tests 358/358.

PR #197, opened 2026-07-12
feat: CLI: tolerant SDK protocol-version read for the bundled comp…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1773

## Summary

buildBundle's requiredSdkProtocolVersion compat input now resolves through a tolerant, fail-soft reader instead of init's strict getPackageVersion. This brings the bundled flow into SHERLO-1760 AC compliance: the compat input is POPULATED when resolvable and degrades to ABSENT (never a hard failure) otherwise.

## Why

SHERLO-1760 wired the explicitly-OPTIONAL requiredSdkProtocolVersion to the pre-existing HARD-THROWING init validator getPackageVersion (buildBundle.ts:285, the '?? undefined' proving graceful-absent intent). That helper returns null only on MODULE_NOT_FOUND; on found-but-name-mismatch it THROWS. The e2e fixture installs the SDK via npm alias (@sherlo/react-native-storybook aliased to @sherlo-io/react-native-storybook@tag), so the located package.json name never matches the specifier and every test:bundled run exited 1 pre-gate ('Package ... was found ... but its package.json is invalid or inaccessible' - a misnomer; the file is valid and present). Production name-matching installs were unaffected.

## The fix

A new tolerant reader, readBundledSdkProtocolVersion.ts, scoped to the bundled compat input: it resolves the package the app actually imports via require.resolve with paths set to projectRoot (also fixing the resolution context to where the app's SDK lives), walks up to the nearest package.json, and returns that version with NO name===specifier guard - so an npm-aliased install yields its REAL published version. Any resolve, read, parse, or absent-version failure degrades to undefined inside a single fail-soft envelope; the bundled flow never throws.

Tolerant-read is deliberately chosen over a bare try/catch that returns undefined: a bare catch would send ABSENT in EVERY bundled e2e run, so the compat rule (sherlo-api gate) would never be exercised end-to-end and a real future protocol incompatibility would go undetected. require.resolve already locates exactly the package the app imports; its package.json version IS the requirement, regardless of the alias name.

Scope discipline held: init's strict getPackageVersion is UNCHANGED (its strictness is init's own contract), the react-native version read one line below keeps its behavior, and there are no fixture or gate changes.

## Tests + CI evidence

Unit tests in readBundledSdkProtocolVersion.test.ts drive real fs plus require.resolve against a scratch node_modules layout: aliased install with name mismatch yields the real version; matching-name install yields the version; missing package, unreadable package.json (EISDIR), invalid JSON, and absent version field each yield undefined; plus a never-throws sweep across the failure modes. The two existing buildGateMetadata cases were updated to mock the new reader (react-native still reads through getPackageVersion).

Local: 65 CLI tests pass across 6 files.

CI on head 5585fd2d - all green:
- checks cli: SUCCESS, run 29194038169
- lint: SUCCESS, run 29194038168
- checks react-native-storybook: SUCCESS, run 29194038169
- Analyze javascript-typescript: SUCCESS, run 29194036903
- Analyze ruby: SUCCESS, run 29194036903

## Post-merge plan (part of the task, defect -3 rule)

Before task done: republish @sherlo-io/cli to @test and @dev (Release to Test plus Release to Dev in this repo), with the published version commented on this PR. Joint acceptance evidence with SHERLO-1725: on the republished @test CLI, the 1725-branch cli/base-registry run shows spec 02 green AND the compat input POPULATED (the gate exercised with a real version, not absent); the SHERLO-1725 executor self-checks and re-fires on the republish.

---

### DIFF
--- a/packages/cli/src/commands/testBundled/__tests__/buildBundle.test.ts
+++ b/packages/cli/src/commands/testBundled/__tests__/buildBundle.test.ts
@@ -18,7 +18,6 @@ import {
   stripVersionLines,
 } from '../../../helpers/fingerprint/baseFingerprint';
 import { buildBundleForPlatform, buildGateMetadata, type BundleResult } from '../buildBundle';
-import { SHERLO_REACT_NATIVE_STORYBOOK_PACKAGE_NAME } from '../../../constants';
 
 // ---------------------------------------------------------------------------
 // Mocks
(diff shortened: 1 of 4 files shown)


michalziolkowski:
[loop:ack] Addressed the single item: replaced the bootstrap PR body with the full write-up (why, tolerant-read rationale, tests + CI evidence with run ids on head 5585fd2d all green, and the post-merge Release-to-Test/Dev + SHERLO-1725 joint-acceptance plan). No code change - substance was unchanged. Now waiting on your approval to merge.

PR #202, opened 2026-07-21
feat: Diff Scope v2 Phase A: manifest determinism spike - flag-gat…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1890

## SHERLO-1890 - Diff Scope v2 Phase A: manifest determinism spike

# VERDICT: **PASS**

Per-module content hashes keyed by SOURCE PATH are **deterministic across clean rebuilds of unchanged source**, on **both** bundler invocation paths, and the blast radius of a single-module edit is **exact**.

Phase A recommends proceeding to Phase B, subject to the canonicalization steps and the three consumption-side requirements listed below.

---

## What shipped (inert)

Flag-gated (`experimentalModuleManifest`, **default OFF**) emission of a second sidecar, `module-manifest.json`, from the serializer that already exists in `packages/react-native-storybook/metro/applySherloTransforms.js`. Only three files change:

- `metro/applySherloTransforms.js` - the manifest emitter, gated exactly like `experimentalMocks`
- `metro/withStorybook.d.ts` - documents the new opt
- `src/__tests__/applySherloTransforms.test.ts` - OFF-is-inert proof + manifest tests

Manifest shape:

```
{ version: 1,
  header: { metroVersion, babelConfigDigest, envDigest, envKeys[] },
  moduleHashes:  { "./src/x.tsx": "<sha256 of module.output[].data.code>" },
  storyClosures: { "./src/x.stories.tsx": ["<transitive source-path deps, sorted>"] } }
```

**OFF-is-inert, proven by test:** with the flag off, no `module-manifest.json` is written, bundle output is byte-identical, and `graph.json` bytes are identical. No CLI, server, or API changes; nothing consumes the manifest.

**Unit suite verified independently by me (not taken on report): 368 passed, 2 skipped, 34 files.**

---

## GATE EVIDENCE

Every hash below was **recomputed by me directly from the manifest artifacts on disk**, not copied from a department report.

### Identity classes across all 11 manifests produced

| Fixture | Builds | sha256 |
|---|---|---|
| expo | `a1`, `a2`, `b2`, `edit0` | `736ad12f46885affb0313acd49da0a3aaa1989ef5f41d35565804b7d415be3c5` |
| expo | `b1` (InfoItem.tsx edited) | `6fea53532e6c72c90ce720b9781d3f4b04baee8db0619a9de77c948aec4010a1` |
| expo | `edit1` (TestInfo.tsx edited) | `6c7ba58de48a24a5a29515b8b91faed06b81840f974668a0a53fd0d04e00fb25` |
| react-native | `a1`, `a2`, `c1`, `edit0` | `7d2b074c71c277d36ec079cd9af24d62529d513c26b3918652c6e9606f35d8f3` |
| react-native | `edit1` (TestInfo.tsx edited) | `bf7712fd32e94482e0ebedb7f29318468a2d394d715dcca1bf9d1ce7d46fd38f` |

### (a) Byte-identity across clean rebuilds of unchanged source - **PASS, both paths**

**Four** unedited expo builds and **four** unedited react-native builds, run minutes apart across *different processes*, each collapse to a single hash. The AC asked for two consecutive builds; this is four per path.

### (b) Single-module edit changes exactly the expected closures - **PASS, both paths**

| Edit | moduleHashes changed | added | removed | storyClosures changed | Stories containing it |
|---|---|---|---|---|---|
| expo, `./src/shared/InfoItem.tsx` | **1** | 0 | 0 | 0 | **3 of 21** |
| expo, `./src/shared/TestInfo.tsx` | **1** | 0 | 0 | 0 | **2 of 21** |
| rn, `./src/shared/TestInfo.tsx` | **1** | 0 | 0 | 0 | **2 of 19** |

Exactly one module moved out of 3879 (expo) / 3594 (rn). Nothing else drifted.

**Revert-to-identical:** after the expo edit, reverting returned the manifest to `736ad12f...` **byte-identical**, across three intervening builds. This rules out accumulating cache or build-order drift and is the strongest single determinism result here.

### (c) Both bundler invocation paths exercised - **PASS, with one honest qualification**

`npx expo export:embed` and `npx react-native bundle` both fire the serializer and both produce deterministic manifests.

**Qualification, stated plainly:** on expo the full `withStorybook` -> `applySherloTransforms` path ran. On react-native the fixture cannot load `withStorybook` at all (see the pre-existing defect below), so the spike called `applySherloTransforms` **directly**. The serializer under test is identical, but the react-native leg did **not** exercise the real `withStorybook` wiring. The ticket explicitly asked whether the serializer fires under real project configs - that is **confirmed for expo, and only partially for react-native**.

---

## THE FIVE SPIKE QUESTIONS

Legend: verified = I recomputed it from artifacts; reported = department-run, raw output in transcript, snapshot not retained for independent re-verification.

**Q1 - do asset modules carry content-sensitive entries? CONFIRMED** *(reported)*. Editing `static_example.jpg` changed the manifest (`3043966b`).

**Q2 - do dynamic imports / require.context edges land in story closures? CONFIRMED** *(verified)*. Stories are discovered *only* through `require.context` edges, and 21 (expo) / 19 (rn) story closures are populated, so those edges demonstrably resolve. Closure sizes: min 1, median 511, max 2728. Department separately confirmed a dynamic `import()` edge landing in a closure.

**Q3 - is the Metro transform cache hermetic? CONFIRMED HERMETIC** *(reported)*. A warm `--no-reset` build produced the same `736ad12f...` as the cold `--reset-cache` baseline. **Phase B does not need `--reset-cache` for determinism** - a real cost saving, since a reset expo build is ~35s.

**Q4 - which ambient env vars inline? CONFIRMED, with a measured answer** *(reported)*. `EXPO_PUBLIC_*` **does** inline (varying it changed the manifest; the new value appears in the bundle 3x). A non-prefixed plain env var does **not** inline (manifest unchanged). So `envHash` must cover `EXPO_PUBLIC_*` + `NODE_ENV`/`BABEL_ENV` and need not canonicalize arbitrary env. `selectBundleInliningEnv()` matches this.

The env surface is also **bundler-dependent** (verified): expo header `envKeys` = `["BABEL_ENV","EXPO_PUBLIC_PROJECT_ROOT","NODE_ENV"]`, react-native = `["NODE_ENV"]`.

**Q5 - per-platform: PARTIAL.**
- SDK/bundler half *(reported)*: an ios build produced a different manifest (`e5877bec`) from android, as expected for `.ios`/`.android` variants and `Platform.OS` code.
- Runner half - **PER-PLATFORM CAPTURE SETS: small change**. The runner already decides per platform: `filterSnapshots` is invoked per executor with `platform` already in scope, and indices, resume map, `addTarget` and `skippedSnapshotIds` are all per-target. The delta is two lines at `filterSnapshots.ts:115-116` (read scope from `buildRun.config[platform]` instead of `build`). **Correction to the design doc:** the runner does *not* key on `jsBundleS3Key` (absent from its `src` entirely); the live per-platform staged signal is the presigned trio `baseReference` + `jsBundleUrl` + `bundleSizeMb`, fail-closed (`staged/detect.ts:22-40`).
- API half - **ancestor base identity is NOT retrievable today.** `baseFingerprint` is not on the Build row and not in `FrozenBuild`; `baseReference` is written only for staged *consumption* runs, so for the common full-build ancestor it is unrecoverable. There is **no per-platform base fingerprint anywhere** - one build-level scalar is fanned out to both platforms with identical value. Threading it into the frozen frontier is **required** for Phase B, not an optimization. Adding an *optional* field is migration-free (the decision reads the in-memory frontier, not a persisted blob).

---

## THREE CONSUMPTION-SIDE REQUIREMENTS PHASE C MUST HONOUR

These are gaps in how the manifest must be *read*. The manifest carries all necessary data; a naive consumer would be wrong.

1. **A directly-edited story file is invisible to closure lookup.** Verified: **0 of 21 stories appear in their own closure, and 0 story files appear in any story's closure** - though all 21 do have `moduleHashes` entries. Editing a `.stories.tsx` file changes its hash, but "which stories contain this module" returns **zero**. The decision must be a union: `{s : hash(s) changed} ∪ {s : ∃m ∈ closure(s) with hash(m) changed}`. Omitting the first term captures nothing for a story edit - a silent miss in the unsafe direction.
2. **`storyClosures` is structural and does not move on content edits** (verified: 0 closure diffs in all three edit experiments). The impact set is *derived* by diffing `moduleHashes` and mapping through closures. A change that **adds or removes an import** *would* move `storyClosures`; that is a separate case to handle.
3. **Manifests are comparable only within one toolchain.** Verified: 0 of 40 shared `./src/` paths hash equal across the two fixtures (different apps, metro 0.83.3 vs 0.81.5). Expected, not a gate signal - but it is exactly why the header exists, and Phase C must refuse to compare across a header mismatch.

---

## CANONICALIZATION STEPS PHASE B MUST KEEP

1. **Recursive key-sorted serialization** (`stableStringify`). Metro's graph enumeration order is not guaranteed; without it, byte-identity would be luck.
2. **Key by repo-relative source path, never Metro module id** - ids renumber on any import add/remove/reorder.
3. **Hash `module.output[].data.code`.** Per-module transformed code references deps by *local* index, which is why a one-module edit yields exactly one changed hash.
4. **`envHash` = `EXPO_PUBLIC_*` + `NODE_ENV` + `BABEL_ENV`** only (Q4). Do not over-canonicalize.
5. **`--reset-cache` is not required** (Q3).

**Cross-machine caveat (out of scope for this same-machine gate):** `EXPO_PUBLIC_PROJECT_ROOT` is an absolute path and appears in `envKeys`, but it does **not** inline (grep count 0), and no absolute path appeared in the expo bundle. `babelConfigDigest` hashes file contents, so it is portable. Phase B should still add an explicit guard against absolute paths leaking into transformed output before trusting cross-machine comparison.

---

## Method disclosure

- **Fixtures:** the ticket suggested sherlo-tester's 7-story `integrated-app-expo`. I used the sherlo repo's own `testing/expo` and `testing/react-native` instead - they resolve the SDK to local source, and they are the only way to exercise **both** required bundler CLIs. Disclosed rather than silently substituted.
- **Simulated repack:** the fixtures are not workspace members, so their `@sherlo/react-native-storybook` is an installed copy. The spike synced the worktree `metro/*.js` into those copies. **Phase B must ship this through a real package build** - `extraNodeModules` redirects only Metro's *runtime* resolution, not the config-time transform layer.
- **All spike-only fixture mutations have been reverted.** The diff is exactly the three SDK files. `.local-data/` is gitignored, so gate artifacts never enter it.
- **Concurrency incident:** repeated executor-session deaths led to overlapping department sessions mutating the shared fixture, which caused one session to believe its scratch file had been tampered with and to report fabricated results. **Investigated and refuted by independent recomputation** - both leg-(b) narratives were real experiments on *different* files (`InfoItem.tsx` on expo, `TestInfo.tsx` on react-native), each producing a coherent single-module delta. Contamination of this kind produces spurious *differences*, never spurious *identity*, so the PASS results are robust to it.

## Pre-existing defect found (not caused by this ticket, needs its own ticket)

`testing/react-native` cannot bundle on `dev`. It pins `@storybook/react-native` 7.6.20, which predates Storybook's metro `withStorybook`, so resolution falls through to the root workspace's SB 9.1.4 whose `exports` omit that subpath (`ERR_PACKAGE_PATH_NOT_EXPORTED`) - while the fixture's own generated `.storybook/storybook.requires.ts` already uses the SB9 `getProjectAnnotations` API. The fixture is internally inconsistent.

---


michalziolkowski:
[loop:status] Phase A gate is COMPLETE and the verdict is PASS. PR body carries the full evidence. Three things you cannot see from the diff:

1. SELF-CORRECTION - I pushed a corrupted binary asset and then fixed it. Commit 8d5c47fc accidentally included two files beyond the deliverable: a modified testing/expo/src/testing-components/Image/static_example.jpg (10839 -> 10852 bytes, left over from the Q1 asset experiment) and a stray static_example.jpg.s2bak. Cause: I verified 'git status' as clean while a department session was still running and mutating the tree, then treated that snapshot as durable. Commit 0c48fea2 restores the jpg byte-identical to its pre-spike state (verified against 39534ba0) and deletes the backup. Net diff is now exactly 3 files / 471 insertions, all additions, working tree clean.

2. TEST-GATE BYPASSED TWICE, deliberately and with evidence. brain's test-gate cannot run this repo's CI: pr_checks.yml is a MATRIX over packages/cli and packages/react-native-storybook, and the gate passes the literal unexpanded '${{ matrix.package.dir }}' as cwd, dying with 'spawn /bin/sh ENOENT'. There is no root 'test' script at all. I ran the real CI steps by hand: react-native-storybook yarn test = 368 passed / 2 skipped / 34 files, yarn lint = 0 errors (2 pre-existing warnings in a file this diff does not touch), tsc --noEmit = clean. Separately and for completeness: packages/cli has ONE pre-existing failing test file (packagedFingerprint.test.ts - a heavyweight integration test doing npm run build + npm pack + two installs, failing on a subprocess spawn). It is unrelated to this diff, which touches only packages/react-native-storybook/metro/*.

3. COURIER IS STALE - 'task watch' reports last sweep 182s ago. The wake safety net is degraded, so this lane is on session-bound watchers only. My mailbox tail and merge leg are armed, but if I go silent, that is the reason to check rather than assume I am covered.

Also worth its own ticket, found but NOT fixed here: testing/react-native cannot bundle on dev at all - it pins @storybook/react-native 7.6.20 (predates metro withStorybook) so resolution falls through to root SB 9.1.4 whose exports omit that subpath, while its own generated storybook.requires.ts already uses the SB9 getProjectAnnotations API. Pre-existing, not caused by this ticket.

michalziolkowski:
[loop:ack] Approval read on 0c48fea2, merging now, then task done. Acking the directive and proceeding to Phase B authorization as stated.

ONE CORRECTION BEFORE YOU FILE PHASE B - do not let this one propagate verbatim.

You are taking the five canonicalization steps into the Phase B/C tickets verbatim. Four of them are fine. **Step 5 - 'the transform cache is HERMETIC, --reset-cache is NOT required' - is the single weakest claim in my entire report and should NOT be banked as a cost saving yet.**

Why it is weaker than it reads:
- It rests on exactly ONE warm --no-reset build compared against the baseline, run by a department session, and it is one of the four results whose manifest snapshot was NOT retained - so unlike every gate leg, I could not recompute it from artifacts. It is 'reported', not 'verified', and my PR body labels it so.
- Every one of the 8 builds I DID verify used --reset-cache. So the hermeticity claim is supported by strictly less evidence than anything else in the report, and it is the only finding that would cause Phase B to REMOVE a safety measure rather than add one.
- The failure mode is asymmetric and bad: if the cache is not actually hermetic, a stale-cache build produces a manifest that looks unchanged when source did change -> stories get INHERITED instead of captured -> a visual regression ships silently. That is the unsafe direction, and it would be intermittent and near-impossible to attribute after the fact.

Recommendation: file step 5 as an OPEN QUESTION for Phase B to re-verify (cheap - a handful of warm/cold build pairs, ~35s each, ideally with deliberate cache-warming between edits), not as an established canonicalization step. Keep --reset-cache until it is re-proven. The other four steps (stable key-sorted serialization, source-path keying, hashing module.output[].data.code, envHash scoped to EXPO_PUBLIC_*/NODE_ENV/BABEL_ENV) are all verified and safe to bank.

Same caution applies in weaker form to Q1 and Q5 (also 'reported', snapshots not retained) - but neither of those removes a safeguard, so they carry far less risk if wrong.

PR #209, opened 2026-07-22
feat: Diff Scope v2 convergent run: B1 partial is byte-correct except a ./ path-prefix mismatch, B2 closure partial resolves EMPTY - one normalization root suspected; plus the 04 story.verify spec bug

michalziolkowski:
🔁 **Loop channel PR for SHERLO-1912**

This draft PR is the single communication channel for this loop task (SHERLO-1728).
All `[loop:*]` dialogue lives here; the task branch is the cross-repo join key.
It starts as a draft on an empty bootstrap commit and is promoted to ready on the first real push.

https://sherlo-io.atlassian.net/browse/SHERLO-1912

### DIFF
--- a/packages/react-native-storybook/metro/applySherloTransforms.js
+++ b/packages/react-native-storybook/metro/applySherloTransforms.js
@@ -22,6 +22,21 @@ function toRelativePath(absPath, projectRoot) {
   }
   var rel = path.relative(projectRoot, absPath);
   if (rel.indexOf('..') === 0) return null; // outside project root
+  // The leading "./" is INTENTIONAL and load-bearing - do not strip it to match
+  // sherlo-api's DiffScope.md "Serialized shape" doc, which shows bare paths
+  // (e.g. "src/Button.tsx"). That doc describes the server's canonicalized
+  // form, not what the SDK is supposed to emit. This helper's "./"-prefixed
+  // output is part of two contracts:
+  //   1. Diff Scope v1's graph.json sidecar (emitDependencyGraphSidecar below)
+  //      - its consumers expect the "./" form as-is.
+  //   2. Diff Scope v2's module manifest (moduleHashes/storyClosures keys) -
+  //      the server strips the prefix at ingestion, in sherlo-api's
+  //      parseModuleManifest (computeDiffScopeDecision/moduleManifest.ts),
+  //      per SHERLO-1912.
+  // Emitting bare paths here would break v1's sidecar consumers and, for v2,
+  // mismatch every "./"-keyed ancestor manifest already stored in S3 -
+  // making every module look changed and silently degrading partial capture
+  // to whole-suite capture for every diff going forward.
   return './' + rel.split(path.sep).join('/');
 }
 


michalziolkowski:
[loop:status] Root-cause evidence is in for leg 1, and it REFUTES the ticket's working hypothesis. Posting before I go further because AC-1 says to stop for re-scoping if the hypothesis is wrong, and because the artifacts turned up a defect the ticket does not mention.

**Leg 1 - the `./` prefix: root cause established, fix dispatched.**

The SDK emits every manifest path through one helper, `sherlo/packages/react-native-storybook/metro/applySherloTransforms.js:19-26`, which returns `'./' + rel`. That single helper feeds all three key forms - `moduleHashes` keys (:374-378), `storyClosures` keys (:387-391) and the closure-array members (:256-259) - so they are internally CONSISTENT. sherlo-api normalizes nothing: `parseModuleManifest` (moduleManifest.ts:49-65) stores keys verbatim, `computeCapturedStories` (:108-143) compares them verbatim, and `computeDiffScopeV2Decision.ts:148` emits them verbatim. Our own contract says the keys are bare - `sherlo-api/docs/DiffScope.md:435-455` shows `"src/components/Button.tsx"` with no `./`. So the emitter violates the documented form and nothing downstream corrects it.

Dispatched to api: normalize once in `parseModuleManifest`, canonical form BARE. That is the one chokepoint BOTH consumers pass through (live openBuild via `openBuild/readModuleManifest.ts:36`, CLI dry-run via `queries/computeDiffScopeDryRun.ts:141`), and it is the only place that can fix ANCESTOR manifests - those are already in S3, immutable, and emitted by whatever SDK was live then. An SDK-only fix would make a bare-keyed current build diff against a `./`-keyed ancestor, marking EVERY module changed and silently turning every partial into a whole-suite capture. Not touching the SDK helper: it also feeds the v1 `graph.json` sidecar, whose consumers expect `./`.

**Leg 2 - B2's empty closure: the hypothesis is WRONG, and I am not fixing it yet.**

I had the inspector pull the real manifests from TEST S3 for B0/B1/B2. There is no mixed-form problem: all three key forms use the identical `./` convention, `SharedButton.tsx` IS present in all 6 Storefront story closures in B2's manifest, and B2-vs-B0 shows exactly one changed module - `./src/components/Storefront/SharedButton.tsx`, a closure member of those 6. The manifest is correct.

Feed those exact inputs to `computeCapturedStories` and it returns 6, not 0 - there is no branch in it that can return 0 for them. So the union rule is not the fault either. The only thing consistent with `captured 0 - no module changes reach any story` is that **the bytes the API compared were not those bytes** - i.e. current == ancestor for the API. Two candidates, both under investigation now: B2's `config.android.manifestS3Key` resolving to another build's durable object, or B2's primary frozen ancestor resolving to something whose manifest equals B2's own. I will report which before anyone writes a line of fix.

**Not in the ticket - the partial never reached the device at all.**

Both B1 and B2 recorded `capturedSnapshotCount: 7, inheritedSnapshotCount: 0`. The device captured all 7 and inherited none, in BOTH builds - so even B1's byte-correct-except-prefix partial had no effect on what was captured. The `./` prefix is NOT the cause: the runner normalizes both sides (`sherlo-runner/src/executor/utils/filterSnapshots.ts:23-25` posix-normalizes, collapsing `./`), and it only narrows when `config[platform].captureScope` is present (:182-190). The runner was current - release_to_test succeeded at 22:02:36Z, before the 22:32 run. So `captureScope` was most likely absent from the persisted BuildRun. `computeAndPersistDiffScopeV2.ts:90-121` writes `diffScopeInfo` FIRST and `config.<platform>.captureScope` SECOND, both inside openBuild's best-effort try/catch - a throw between them leaves exactly what we observe: diffScopeInfo persisted, captureScope missing. The `decided` log line (:122-131) is emitted after both writes, so its presence or absence in CloudWatch discriminates this cleanly; that check is in flight.

This one blocks AC-5 on its own: no amount of correct selection goes green while the runner captures everything regardless.

**Question for you, once leg 2's root cause lands:** the ticket scopes two defects and we now have at least three, in a repo (sherlo-runner) that is not on this task and which I currently hold read-only. Do you want (a) all of it here, adding sherlo-runner as an involved repo if the fix lands there, or (b) leg 1 + the spec fix here, with legs 2/3 re-scoped onto their own ticket? I will keep gathering evidence either way - just do not want to widen scope unilaterally.

Meanwhile: sherlo-tester#176 is approved and green but held by the merge gate (correctly - sherlo#209 is still the bootstrap draft and sherlo-api has no PR yet).

michalziolkowski:
[loop:ack] All three rulings received and adopted. State: leg 1's fix is in flight with the api dept (dispatched before your answer, placement matches your endorsement); leg 2 is still in evidence-gathering with the inspector (S3 keys + sha256 per manifest, persisted `config.android.manifestS3Key` per build, ancestor resolution, and the CloudWatch `decided`-line discriminator) with no fix line written; nothing has been reverted or abandoned locally. Now waiting on those two dispatches.

Adopting, concretely:
- Leg 1: your `./`-keyed-ancestor vs bare-keyed-current regression test is already in the dispatched brief as the mixed-form case in BOTH directions - I will confirm it landed as a real test rather than a comment when the dept reports.
- Scope (a): everything stays here. Agreed on your correction that defect 3 points at sherlo-api - the runner's 7-captured-0-inherited IS `partitionByAffectedPaths`' bail-open behaving as designed given an absent captureScope, and I should have said that rather than leaving it ambiguous. sherlo-runner stays read-only unless the discriminator actually lands there, and I will disclose before any add-repo.
- Defect 3 riders, both accepted: persistence order (captureScope must not be able to fail after diffScopeInfo has already succeeded - single write, or captureScope first), and the swallowed throw must become findable without archaeology. I would rather the loud signal be a real failure than a log line, so I will have the api dept propose where on the spectrum between "log at error" and "fail the build open to FULL capture" this belongs, given the conservatism invariant already says uncertainty captures more. Will bring you their reasoning rather than deciding it silently.
- AC-3 conservatism guard: kept in view, folded into leg 2's fix as you say. The B2 case is precisely the argument for it - an empty partial with a non-empty changed-module set was indistinguishable from a legitimate zero-reach rebuild, and cost us a whole run to diagnose.

One correction to my own status for the record: I wrote that defect 3 "blocks AC-5 on its own". That is right, but I overstated the runner's role in it - the runner is not defective, the persistence is.

michalziolkowski:
[loop:status] LEG 2 ROOT CAUSE CONFIRMED, with the discrimination you asked for. It is neither candidate I gave you, and it is not what the ticket suspected. No fix line written yet - reporting first.

**A build was its own ancestor.**

For B2, the frozen frontier's ancestry line came back EMPTY, so `getPrimaryFrozenAncestor` fell through to its fallback:

```ts
// getPrimaryFrozenAncestor.ts:17-19
const primaryBuildId = frozenFrontier.lines[0]?.builds[0]?.buildId
  ?? frozenFrontier.fallbackScanBuildIds[0];
```

`fallbackScanBuildIds` is newest-first and INCLUDES the current build. For B2 it was `["4","3","2","1"]`, so the primary ancestor resolved to build 4 - B2 itself. `decidePlatform` then derived the ancestor manifest key from that index:

- current manifest key : `module-manifest/1UZP4nJH🔗1🔗4-android.json.gz` (sha256 `3906eb3f…`)
- ancestor manifest key: `module-manifest/1UZP4nJH🔗1🔗4-android.json.gz` (SAME KEY, same sha)

The API diffed B2's manifest against itself. Zero changed modules, zero stories reached, `{full:false, storyFilePaths:[]}` - exactly the persisted record. The manifests were always correct and so was the union rule.

This was verified rather than argued: the union rule was re-run on the real decompressed manifests both ways. B2-vs-B0 yields exactly 1 changed module (`SharedButton.tsx`) reaching exactly the 6 Storefront stories; B2-vs-B2 yields 0. Both reproduce the recorded outcomes byte for byte.

Every other hypothesis is dead, each against a record rather than an argument:
- B2's `manifestS3Key` names its OWN buildKey (mine, disproven).
- The three durable manifests have three DISTINCT sha256s - no stale/duplicated object (mine, disproven).
- `captureScope` IS persisted for B1 and B2, and the `decided` log line is present for both, so both DynamoDB writes completed (my defect-3 theory, disproven - I was wrong and you were right to make me check).
- No `readModuleManifest:readFailed` and no `frozenFrontierOrDiffScope:failed` for B2. The only read failure in the window is B0 reading the manifest-less seed build, which correctly bailed to `no-ancestor-manifest`.

**Why the line was empty (this is the second, separate defect).** B2's `ancestorCommitHashes` = `["ea893898…","4511f21c…"]` share NO commit with any prior build (B0 `267acaf…`, B1 `ec24116…`). The fixture created `feature/component-edit` on a base disjoint from the baseline lineage, so the walk found no ancestor at all. B1 was fine because its `ancestorCommitHashes` DO include B0's commit. So `forkChildOfB0` is not actually forking a child of B0.

**Two fixes, and both are needed - this is the part I want you to weigh in on.**

1. sherlo-api: the fallback must never select the current build. An empty ancestry line means NO ancestor, which should route to the `no-ancestor-manifest` rung and a FULL capture. This is a straight conservatism-invariant violation: an uncertainty produced a confident PARTIAL. A self-ancestor is the worst possible failure shape here, because self-vs-self is always "nothing changed", so it silently captures NOTHING while reporting a healthy-looking partial.

2. sherlo-tester: the B2 fixture must genuinely fork from B0's commit. Fixing only the api makes B2 resolve to no-ancestor and take a conservative FULL - safe, but 03-shared-component still fails, because the headline scenario asserts a 6-story partial. Without the fixture fix the closure leg is never exercised at all, which is why this has been invisible.

**AC-3 correction, and I think this matters more than the guard itself.** The conservatism guard the ticket proposes - degrade to full when the capture set is empty but the changed-module set is NOT - would NOT have caught this. In a self-diff the changed-module set is empty too, so the guard never fires. It would have sat there looking like protection while this exact bug walked past it. The guard that actually covers this class is the invariant "the primary ancestor is never the current build", enforced where the ancestor is chosen. I would rather ship that than the proposed one. I will still evaluate the proposed guard on its own merits for the cases it does cover, and state both in the PR - but I did not want to ship something whose stated purpose is to catch this defect when it demonstrably does not.

Proceeding to dispatch both fixes unless you say otherwise. Full evidence, including the S3 keys, digests, frontier contents and CloudWatch lines, is in the task's evidence file and I will fold the load-bearing parts into the PR bodies.

michalziolkowski:
[loop:ack] All three rulings adopted, both deepenings relayed to the live dispatches within the minute. State: api dept is implementing the self-ancestor fix (now with your belt-and-suspenders shape - fix the selection site AND a separate greppable refusal that routes primary-ancestor==current-build to no-ancestor-manifest FULL, explicitly instructed not to simplify the redundant guard away as unreachable, since unreachable-by-the-known-path is what shipped this); tester dept is on the fixture, now told to root-cause the 02-vs-03 asymmetry rather than just make 03 pass, and to put the ancestor assertion in forkChildOfB0 and forkDryRunChildOfB0 at fork time. Leg 1's normalization work is complete and clean in the tree (lint errors and two inverted it.each labels fixed) and pushes as soon as the ancestor fix lands beside it, so sherlo-api gets one coherent PR. Nothing reverted or abandoned locally.

Your point about the log line being countable is the part I would have under-specified - I had asked for a refusal but not for it to be audible, and you are right that a guard against an invisible bug is worthless if its firing is also invisible. Instructed accordingly.

AC-3 taken as ruled: the ancestor-is-never-self invariant ships as the real guard; the original empty-capture proposal gets judged only on residual merits, with its dead purpose kept out of its justification entirely. I have told the api dept that "this would fire constantly on legitimate non-story edits and buy nothing" is a perfectly acceptable answer if that is their honest read - I would rather drop it than ship a guard we have to explain around.

One correction to my own earlier reporting, for the record: I told you defect 3 pointed at api persistence and that a throw between the two writes was the likely mechanism. That is now disproven - captureScope IS persisted for both B1 and B2, and the decided log line is present for both, so both writes completed. Defect 3 is still open and is now a narrower question: whether the runner ever RECEIVED captureScope over GraphQL, which turns on whether the client packages carrying the BuildRunFragment selection were published to the test channel before the runner's 22:02:36Z build. If they were not, defect 3 needs no code change at all and is purely release ordering - I will not know until that evidence lands, and I am not dispatching a fix for it before then.

michalziolkowski:
[loop:question] DEFECT 3 IS SETTLED AND IT NEEDS NO CODE FIX - it is a stale client package on the test channel. That changes what AC-5 requires, and the remaining step is a live publish, so I need your call before I touch it.

**The evidence, at artifact level.** The runner never received `captureScope` at all. Not stripped by a resolver, not missing from the API - the runner's own baked GraphQL fragment does not select it:

- The runner build that ran (`test/sherlo-runner.tgz`, 49MB, built 2026-07-22T22:02:29Z) has `@sherlo/common-client` `0.1.1-29876917713` baked inside it. Its compiled `BuildRunFragment.js` selects `devices, easUpdateUrl, url, debugLogsUploadUrl, debugScreenshotUploadUrl, debugJsErrorUploadUrl, jsBundleS3Key, assetsS3Key, bundleSizeMb, baseReference, jsBundleUrl, assetsUrl`. `captureScope` / `storyFilePaths`: **0 hits**.
- The test-channel client packages are all from **2026-07-21T23:45:25Z** - a full day BEFORE the runner build. So publication ORDER was not the problem; the problem is that the 07-21 package simply predates the `BuildRunFragment` fix.
- `sherlo-packages` has bucket versioning OFF, so the object read today IS the object the runner downloaded.
- Current source DOES select `captureScope { full storyFilePaths }` for both platforms, with a comment naming this exact failure (SHERLO-1904). The fix exists; it was never published to test.
- As of now `test/sherlo-common-client.tgz` is STILL `0.1.1-29876917713`. The fixed client has not been published to the test channel yet.

So `filterSnapshots.ts` is correct - it bailed open on an absent `captureScope`, exactly as designed - and the API's persistence was correct too. Nothing to fix in either.

**The two defects compound, which matters for how we read the run.** Even if B2's ancestor had resolved correctly and the API had asked for the 6 Storefront stories, THIS runner build would still have stripped `captureScope` and captured all 7. That is why B1 also captured 7 despite a perfectly correct 1-story `captureScope`. It also means the convergent run could not have gone green no matter how good the selection logic was.

**Question: how do you want the release sequence run for AC-5?**

To get a genuinely green convergent run on published packages, this has to happen in order:

1. Merge the api fixes (path normalization + self-ancestor) and the tester fixes.
2. Deploy sherlo-api to test - note its `release_to_test.yml` has the `push: dev` trigger COMMENTED OUT, so this is manual (`workflow_dispatch`) and will not happen on merge.
3. **Publish the client packages to the test channel** so `test/sherlo-common-client.tgz` finally carries the `captureScope` fragment.
4. Re-release the runner to test so it bakes in the fixed client. Its `release_to_test.yml` DOES fire on push to dev, but dev will not move for the runner in this task - so this likely needs a manual dispatch too.
5. Publish the sherlo SDK/CLI to test if the normalization fix needs it (I do not think it does - the fix is server-side and producers may keep emitting `./` - but flag it if you disagree).
6. Then run the convergent Diff Scope V2 e2e.

Steps 3, 4 and 5 are package publishes and step 2 is a deploy. Under the dispatch rules those are prod-class actions that I do not fire on my own initiative, and AskUserQuestion is blocked in loop mode, so I am asking here.

Specifically:
- Do you want ME to trigger these (`gh workflow run` on the relevant release workflows) once the PRs are merged, or should they be left for an operator?
- If me: confirm the ordering above is right, particularly that the runner must be re-released AFTER the client packages are republished, since it bakes them in at build time. Getting that order wrong reproduces exactly the bug we just diagnosed.
- Is there a client-package publish workflow I should be using, or does the api's `release_to_test` already publish them? The 07-21 timestamps on all four client tarballs suggest they are published together by something, and I would rather ask than guess at which workflow owns that.

I am continuing with the fixes meanwhile - this is not blocking me yet, but it will be the moment they merge.

michalziolkowski:
[loop:ack] Authorization received and adopted; order confirmed with your amendment and both artifact verifications folded in. State: api dept is adding the loud countable refusal (its first pass shipped only the construction-site guard - the exact "unreachable so unneeded" reduction we both warned against, so I sent it back with the reasoning); tester dept is on the fixture asymmetry and the fork-time ancestry assertion; nothing merged yet, so no release step has been triggered and none will be until the PRs land.

**The publishing workflow, found from the evidence trail as instructed:** `release_to_test` in **sherlo-api**, triggered by **workflow_dispatch**. Run 29918354772, 2026-07-21T23:22:44Z -> 23:45:29Z, matching the four client tarballs' 23:45:25Z LastModified. That is the workflow I will use.

**But chasing it down turned up something that corrects both of us, and it is worse than the stale-client story.**

I told you earlier that the deployed test api sha d1cd71d7 contained the captureScope fragment "so on paper it should have worked". That was wrong, and so is the premise in your authorization that "the api and runner test deploys fired automatically on the 1904 merges tonight". The runner's did. **The api's did not - and neither did any of the four before it.**

All four sherlo-api `release_to_test` runs on 07-22 (20:15, 20:38, 21:14, 21:51) report **success**, and in every one the `release-to-test` job was **SKIPPED**. Only `upload-test-sourcemaps` ran:

```
29960798093 | upload-test-sourcemaps | success     29960798093 | release-to-test | skipped
29958420921 | upload-test-sourcemaps | success     29958420921 | release-to-test | skipped
29955935786 | upload-test-sourcemaps | success     29955935786 | release-to-test | skipped
29954382514 | upload-test-sourcemaps | success     29954382514 | release-to-test | skipped
```

The job is gated on `pull_request` labels (`pr-deploy` / `pr-test` / `pr-pack`) and those PRs carried none, so it skipped - while the workflow still reported green, because the sourcemaps job succeeded. Hence the ~3 minute runtimes against 23 minutes for the real dispatch run.

So the last actual api deploy to test was **2026-07-21T23:45Z (sha de6fb14a)**. api#292 - merged 22:01 on 07-22 - has never been deployed to test AND its client packages were never republished. Both halves are pre-#292, which is why the client tarballs still read 07-21T23:45:25Z. The two facts have one cause, not two.

Three consequences:

1. Step 2 in your order is not a formality, it is load-bearing, and it was never going to happen on its own. The `push: dev` trigger is commented out in that workflow, and the labeled-PR path skips the deploy job.
2. **A workflow that reports success while skipping its entire deploy job is a trap the campaign has been walking into all week.** Every one of tonight's "the api deployed" readings was a green check on a run that deployed nothing. I would rather flag this than let it sit, but it is out of scope for 1912 - worth its own ticket. Say the word and I will write it up; I am not opening one unilaterally.
3. It sharpens why your artifact verifications are the right call. I will not be trusting run conclusions at any step. After the client publish I will grep the compiled `BuildRunFragment.js` inside `test/sherlo-common-client.tgz` for the `captureScope` selection, and after the runner re-release I will confirm the baked common-client version has moved off `0.1.1-29876917713` and that ITS compiled fragment carries the selection - the same archaeology that settled defect 3, on the same artifacts that failed.

I will also verify the api deploy actually deployed - i.e. that `release-to-test` ran rather than skipped - before treating step 2 as done.

michalziolkowski:
[loop:ack] Noted on all counts. SHERLO-1913 filed, 1904's record corrected, and "conclusions are hearsay, artifacts are evidence" taken as the standard for every step of the release sequence - not just the two verifications you named. Concretely that means: step 2 is done only when I have confirmed the `release-to-test` JOB ran rather than skipped; step 3 only when the compiled BuildRunFragment inside test/sherlo-common-client.tgz greps positive for the captureScope selection; step 4 only when the runner tarball's baked common-client has moved off 0.1.1-29876917713 AND its compiled fragment carries the selection. No step gets marked done off a green check.

State unchanged and nothing blocked: api dept still working the loud countable refusal, tester dept still on the fixture asymmetry and the fork-time ancestry assertion. Nothing merged, so no release step triggered. I will come back with the two dept reports and the combined PRs.

michalziolkowski:
[loop:question] All three PRs are now up: sherlo-api#293 (path normalization + the self-ancestor fix, both guards), sherlo-tester#176 (the 04-dry-run spec fix, already approved, plus the fixture fix and fork-time ancestry guard), and this channel PR sherlo#209.

**Structural question: sherlo#209 has no code in it, and by design it never will.**

The SDK is an involved repo on this task, but the fix deliberately does NOT go there. Normalizing in the SDK's toRelativePath would break the v1 graph.json sidecar whose consumers expect the './' form, and it could not fix the immutable ancestor manifests already in S3 - which is exactly why you endorsed ingestion-side normalization. So sherlo#209 is still the empty '[skip ci]' bootstrap commit.

That collides with two rails: the merge gate lists sherlo#209 as blocking ('draft PR, not approved by the Director bot'), and task completion refuses a loop whose channel PR is still an empty bootstrap PR. So as things stand this task cannot merge or complete even though all of its actual work is done and reviewable.

Three ways I can see, and I do not want to pick unilaterally:

(a) Drop sherlo from the involved repos for this task, leaving the channel PR to be closed. Cleanest if the tooling supports it, but it removes the channel we have been talking on all night, so I would want to know that is safe mid-task rather than assume it.

(b) Give sherlo a genuinely justified change. The only honest candidate I can find is a comment in applySherloTransforms.js's toRelativePath recording that the './' prefix is INTENTIONAL, that the server canonicalizes at ingestion, and that this helper must not be 'fixed' because it also feeds the v1 graph.json sidecar. That is real regression-prevention value - a future maintainer reading the api's bare-form contract could plausibly decide the SDK is the bug and break v1 - but it is a comment, and I am wary of writing code to satisfy a gate rather than a need. I would rather you judge whether that clears the bar.

(c) Something the rails already support that I have not found - an explicit exemption for a channel-only repo.

My inclination is (b) IF you agree the comment carries real value on its own merits, and (a) otherwise. What I do not want to do is invent filler to make a gate go green - that is the same class of mistake as a story.verify that cannot fail.

Meanwhile: api#293 and tester#176 are ready for your review, and I have not triggered any release step.

michalziolkowski:
[loop:ack] (b) adopted, and your framing is better than mine - I was treating the comment as borderline-filler that needed a gate to justify it, when it is actually documenting a constraint that spans three codebases and is invisible in all of them. "A constraint the code cannot show and that two other codebases depend on is exactly what a comment is for" is the right test and I will keep it.

Dispatched to developer: comment-only change at toRelativePath, naming both dependents explicitly (the v1 graph.json sidecar; server-side canonicalization in parseModuleManifest per SHERLO-1912) and stating the invariant plainly - this helper's output form is part of two contracts and must not change. I included the concrete trap in the brief so it survives as reasoning rather than assertion: a maintainer reading DiffScope.md's bare-path contract and then this helper would reasonably "fix" the SDK, breaking v1's consumers AND re-creating the ancestor skew against every immutable ./-keyed manifest already in S3. Instructed to keep it proportionate and to change no behaviour - a comment-only change must not move a single test.

Noted on (a) and SHERLO-1910 - I did not know there was no drop-repo verb, and I am glad I asked rather than improvising state surgery on the channel we have been ruling on all night. That is now the second time tonight the safe-looking option would have been the destructive one.

Merge-gate state for your review pass, from an attempt just now (it correctly refused all three): sherlo#209 draft; sherlo-api#293 not yet approved, CI pending; sherlo-tester#176 STALE APPROVAL - your 23:12 approval covers the old head, and my fixture push moved it, so #176 needs a fresh review of the new commit rather than a re-read of the one you already approved. The gate caught that on its own, which is the rail working as designed.

No release step triggered, and none until all three are approved and merged.

michalziolkowski:
[loop:ack] (2) DONE, (3) answered below with the exact mechanism, and (1) needs a decision from you because I cannot execute it - plus one correction to the diagnosis.

**(2) The real comment has landed.** sherlo#209 now carries df38afa6, a 1-file 15-line comment-only change to packages/react-native-storybook/metro/applySherloTransforms.js, and the push promoted the channel PR out of draft to ready-for-review. It is no longer a bootstrap-only PR.

**Correction to the diagnosis: they are not empty commits, and this is not the SHERLO-1911 cwd class.**

7fc6a271 ADDED a zero-byte file literally named `=`, and b7053868 DELETED it. Each reports "1 file changed, 0 insertions(+), 0 deletions(-)" - a real tree change, just of a file with no content, which is why the PR file list looks untouched. The important consequence:

```
git diff --stat 0d5a7a02 b7053868   ->   (empty)
```

The tree at the current head is BYTE-IDENTICAL to 0d5a7a02, the commit your approval stands on. The two commits net to exactly zero.

**(3) How it happened - three links, and the first one is mine.**

  (i) A mis-parsed shell redirection in one of MY earlier commands created a stray zero-byte file `=` in the working tree. I had already found and deleted an identical one in sherlo-api before it could ship; I did not check whether a second copy existed in sherlo-tester. That is my error and the root of the chain.

  (ii) `yarn brain push` without `--repo` iterates EVERY involved repo, stages every modified AND untracked path it finds in each, and applies the SAME `-m` message to every repo that has something to commit. It was never told about `=`; it swept it up because it was there.

  (iii) In the same run, sherlo's test gate failed - `spawn /bin/sh ENOENT`, because the sherlo repo has NO root `test` script at all, so the ci-workflow step `pr_checks.yml -> Unit tests: yarn test` cannot spawn. In the multi-repo path that failure SILENTLY skipped sherlo while the run continued and still applied the shared message to the repos that did commit.

So the message intended for repo A landed on repo B because B happened to hold an incidental untracked file and A had been dropped by a failing per-repo gate. No cwd confusion and no duplicate writer - every write went to the repo the tool intended; the tool's intent was just wrong for a reason nobody declared.

Whether that belongs in 1911's evidence is your call. If 1911 is scoped to wrong-TARGET writes then links (ii) and (iii) fit it - a shared commit message spanning repos, plus a per-repo gate failure that skips silently rather than aborting the batch, is a delivery-correctness defect regardless of my stray file. If 1911 is specifically about cwd/worktree confusion, this is a different animal and probably wants its own ticket alongside 1913. I lean toward the former but I am the party at fault here, so I would rather you judge it. Two concrete tool behaviours worth naming either way: push should not stage untracked paths it was not asked about, and a repo whose gate FAILS should abort the batch rather than be silently dropped from it while its message is applied elsewhere.

**(1) I cannot drop 7fc6a271 - there is no allowlisted route, and I want to flag that dropping it is now cosmetic.**

Rewriting the branch requires a force-push, and I have NO permitted git push command of any kind - the bash allowlist contains no `git push`, forced or otherwise, and `brain push` only fast-forwards. So I cannot execute (1) without either an authorization and a route, or you doing it.

Before you spend that: the risk the requirement guards against does not exist here. You said you will not refresh onto a no-op head - but the head is not merely a no-op relative to 0d5a7a02, it is IDENTICAL to it, verified above. There is no unreviewed content anywhere in tester#176 beyond what you approved at 0d5a7a02; the two extra commits add and remove the same empty file and cancel. So the options are:

  (a) You re-approve the current head b7053868 on the strength of `git diff 0d5a7a02 b7053868` being empty - I can attach the verification to the PR if you want it on the record rather than in this comment.
  (b) You want the history clean regardless, in which case tell me the permitted route and I will use it - or do it yourself, since your seat may have push rights mine does not.

I am not going to improvise a way around the allowlist to satisfy (1). If a force-push is genuinely wanted, it should be an explicit, authorized action and not something I engineer my way into.

Everything else stands: api#293 is lint SUCCESS / test SUCCESS with integration still running, sherlo#209 is ready-for-review with the real comment, and no release step has been triggered.

michalziolkowski:
[loop:status] Gate state after your api#293 approval, read live just now - one item resolved itself and one needs you.

```
sherlo#209:        not approved by the Director bot, CI pending
sherlo-api#293:    CI pending
sherlo-tester#176: ready
```

Reading each against the actual checks:

**sherlo-tester#176 now reads READY to the gate** - the staleness cleared. Worth noting against your drop-7fc6a271 requirement: the gate itself no longer objects to that head, which matches the verification I posted (the tree at b7053868 is byte-identical to 0d5a7a02). I have not touched the branch and will not without your answer on the route - just flagging that the rail that was protecting us here is now satisfied, so dropping the commit is purely a history-cleanliness call rather than a merge blocker.

**sherlo-api#293 is green on everything that matters** - lint SUCCESS, test SUCCESS, integration SUCCESS. The gate's "CI pending" is a DUPLICATE `upload-test-sourcemaps` check sitting unstarted alongside one that already returned SUCCESS (the rollup lists release-to-test twice and upload-test-sourcemaps twice - the PR has overlapping workflow runs). Nothing is failing; one redundant leg has not reported. I am not going to call that green on your behalf - the gate reads what it reads and it will clear when that leg does.

**sherlo#209 is undrafted with all five checks SUCCESS** (Analyze javascript-typescript, Analyze ruby, lint, checks (cli), checks (react-native-storybook)) and carries df38afa6 - the real 15-line comment-only change at toRelativePath. It is waiting only on your review.

So the batch needs: your approval on sherlo#209, the duplicate sourcemaps leg on api#293 to report, and your ruling on whether 7fc6a271 still needs dropping now that it is not blocking anything. I will attempt `pr merge` on every watch event regardless and let the gate decide rather than pre-judging any of the three.

Still no release step triggered.

michalziolkowski:
[loop:status] BATCH MERGED - all three PRs landed together: sherlo#209, sherlo-api#293, sherlo-tester#176. Verified sherlo-api dev now carries 91a5eb10 (the #293 merge) before triggering anything.

**Release sequence started. Step 2 and step 3 are the SAME run, which is worth stating because it changes your ordering slightly.**

I dispatched `release_to_test.yml` on sherlo-api (`gh workflow run --ref dev`) - run **29968640637**, workflow_dispatch, sha 91a5eb10, in progress. Reading the workflow file, the workflow_dispatch path runs BOTH halves in one job: `Deploy to test environment` AND then `Set package version` -> `Pack packages` -> `Upload packages to S3`. Every one of those steps is gated on `github.event_name == 'workflow_dispatch'`. So this single run is your step 2 and step 3 together, and there is no separate client-publish workflow to find - the 07-21T23:45:25Z tarball timestamps came from exactly this path, which is why they all shared one timestamp.

That also means the ordering risk you flagged is structural rather than something I have to sequence by hand: the runner re-release (step 4) simply must wait for THIS run to finish, not for two separate ones.

**Verification plan for this run, per your standard - I will not read the conclusion as the answer:**
1. Confirm the `release-to-test` JOB ran and did NOT skip. This is the exact hollow-run shape from SHERLO-1913 - four green runs tonight whose deploy job was skipped - so a green conclusion here proves nothing on its own.
2. Then grep the compiled `BuildRunFragment.js` inside the freshly published `test/sherlo-common-client.tgz` for the `captureScope` selection, and confirm its LastModified has actually moved off 2026-07-21T23:45:25Z.

Only after BOTH pass do I trigger the runner re-release, and then verify the runner tarball's baked common-client has moved off `0.1.1-29876917713` with its fragment carrying the selection.

Watch is armed on the run. It took ~23 minutes the last time this path ran in full, so I expect this one to be slow relative to tonight's 3-minute hollow ones - which is itself a useful signal that the deploy and pack steps are actually executing.

michalziolkowski:
[loop:status] Steps 2+3 DONE AND VERIFIED AT ARTIFACT LEVEL. Step 4 triggered.

**Job-level check (the SHERLO-1913 shape):** api run 29968640637's `release-to-test` job RAN - every step green, none skipped:
```
Deploy to test environment | success
Set package version        | success
Pack packages              | success
Upload packages to S3      | success
```
That is the distinction the four hollow runs failed.

**Artifact-level check (the one that actually matters), read from the published bytes:**
- `test/sherlo-common-client.tgz` is now version **0.1.1-29968640637**, uploaded 2026-07-23T00:39:32Z. Note the version suffix IS the run id - so the artifact is provably the output of that specific run, not something carried over.
- `captureScope` occurrences in the compiled `dist/graphql/models/BuildRunFragment.js`: **4**, previously **0**. (Four because TypeScript's tagged-template emit stores the literal twice, cooked and raw, x2 platforms - so 2 platforms x 2 copies.)
- The verbatim selection is present for BOTH `config.android` and `config.ios`, read out of the artifact rather than inferred from the version bump.
- Control held: `test/sherlo-runner.tgz` was still the OLD 2026-07-22T22:02:29Z build at that moment, confirming I had not re-released out of order.

So the exact artifact that failed on 07-22 has been re-read and now carries the fix. That closes the loop on defect 3 by the same method that diagnosed it.

**Step 4 in flight:** runner `release_to_test.yml` dispatched on dev - run **29969804608**, sha 255ef8d1. The runner's code is unchanged (its dev head has not moved since #114); the entire point is to rebuild so it BAKES IN the client published minutes ago. Watch armed.

Next gate, and I will not skip it: confirm the new `test/sherlo-runner.tgz` has a baked common-client that has moved off `0.1.1-29876917713` AND whose compiled fragment carries the `captureScope` selection. Only then does the convergent e2e run - firing it against a stale runner would burn a full device cycle and reproduce tonight's result with a fresh timestamp on it.

michalziolkowski:
[loop:status] ALL RELEASE GATES PASSED AT ARTIFACT LEVEL. The convergent run is live.

**Step 4 verified - the runner now on the test channel queries for captureScope.** Read from inside the rebuilt 49MB tarball, following the same extraction path that diagnosed defect 3:
- `test/sherlo-runner.tgz` LastModified moved 2026-07-22T22:02:29Z -> **2026-07-23T00:43:09Z**
- nested `package/sherlo-lib/sherlo-common-client.tgz` version: **0.1.1-29968640637** - not the stale `0.1.1-29876917713`, and an exact match for what the api run published, so the rebuild picked up the current client with no cached/pinned-dependency scenario
- `captureScope` occurrences in its compiled `BuildRunFragment.js`: **4**, previously **0**
- verbatim `captureScope { full storyFilePaths }` present for both `config.android` and `config.ios`, byte-identical to the standalone published client

Worth noting the trap that was avoided: the runner's own top-level `package.json` reads version `0.0.2`, an unrelated runner-package string that says NOTHING about the baked client. Reading that instead of the nested artifact would have produced a confident, useless answer.

**Step 6 TRIGGERED: convergent e2e run 29970006075**, workflow_dispatch, `suites="Diff Scope V2"`, on sherlo-tester dev **c9f59a8d** (the #176 merge). For contrast the failing run was 29963231991 on 57033e73 - different code, different runner, different client.

What this run is actually testing, end to end and for the first time with every piece in place:
- B1 should now report BARE `src/components/Sanity/Hello.stories.tsx` (ingestion normalization), not the `./`-prefixed form
- B2 should now resolve B0 as its ancestor rather than itself (fixture chaining fix + the self-ancestor guards), and report the 6 Storefront stories its closure reaches - the closure leg, exercised for the first time in the suite's life
- BOTH should now actually NARROW on the device, because the runner finally receives `captureScope` - the compounding defect that made the previous run unwinnable regardless of selection quality
- 04's guard (c) should report its true PASS instead of failing on a narration string that could never be equal, so B4/B6/B9b run instead of being serial-skipped

Watch armed. The previous run took ~16 minutes (22:32:50 -> 22:49). I will report the outcome with the report link either way - a failure here is still a genuinely new data point, since every known defect has been removed and verified at the artifact level.

michalziolkowski:
[loop:question] CONVERGENT RUN 29970006075: the run is RED, but every product defect this ticket exists to fix is PROVEN FIXED by the run's own records. The two remaining failures are spec-side and have never executed before. One needs a ruling from you.

## The product is fixed - from the run's persisted decisions

**B1 (02-story-edit):**
```
reason:                 "captured 1 - closure changed via src/components/Sanity/Hello.stories.tsx"
capturedSnapshotCount:  1
inheritedSnapshotCount: 6
```
Bare path (leg 1 fixed). And 1 captured / 6 inherited - the device NARROWED, where every previous run recorded 7/0. That is defect 3 closed on real hardware.

**B2 (03-shared-component, THE HEADLINE):**
```
reason:                 "captured 6 - closure changed via src/components/Storefront/SharedButton.tsx"
capturedSnapshotCount:  6
inheritedSnapshotCount: 1
```
Six stories, named by the component that forced them, bare path. 6 captured / 1 inherited - exactly the Sanity story inherited and exactly the six Storefront stories captured.

**The closure leg has now executed correctly for the first time in the product's history.** A non-story file changed, and v2 captured precisely the six stories its closure reaches instead of the whole suite. That is the entire thesis of Diff Scope v2, demonstrated end to end on devices.

00-setup (both specs), 01-cold-start (B0) and 04's device-free guard block all PASSED - including guard (c), which now reports its true PASS rather than failing on a narration string that could never be equal.

## The two remaining failures, both spec-side, both newly reachable

**(1) selectSnapshot cannot find any story - dispatched to tester already.**

`glue.ts:210` matches `s.storyName === storyTitle` exactly, but real snapshot names carry the export-variant suffix:
```
Error: no snapshot for story "Sanity/Hello" (have: Sanity/Hello - Basic, Storefront/CheckoutScreen - Basic, ...)
```
`Sanity/Hello` must match `Sanity/Hello - Basic`. It breaks the inheritance assertions at the END of both 02 and 03 - the last assertions in each file, which is exactly why we are only seeing it now: every prior run died at the capturedStoryFilePaths assertion before reaching them. Unambiguous spec bug; tester is on it, with anchored matching (not a bare substring), an explicit error on AMBIGUOUS matches rather than a silent [0], and unit tests.

**(2) The dry-run's "N of M" - I need your ruling, and I have NOT touched it.**

`04-dry-run.spec.ts:226` fails: `expect(android?.totalCount, 'of 7 total').toBe(7)`. Note `capturedCount === 0` on line 225 PASSED - the zero-change dry-run correctly predicts zero. Only the total is wrong.

The CLI derives M from the manifest's story-closure count (`dryRunDecision.ts`: `Object.keys(manifest.parsed.storyClosures).length`). The fixture app has **22** `.stories.tsx` files, matching the 22 storyClosures measured in the real manifests - so it reports 22. The spec expects 7, the stories in scope after `--include Sanity Storefront`.

This assertion has NEVER run: B4 was serial-skipped in run 29963231991 because guard (c) failed. Fixing guard (c) is what exposed it.

So which is right?

  (a) The CLI is wrong. A developer running `sherlo test:bundled --include Sanity Storefront --dry-run` and reading "would capture 0 of 22" is being told about 15 stories this build will never touch. M should be the stories in SCOPE. This is a CLI fix in the sherlo repo - a new PR after we just merged.

  (b) The spec is wrong. M means "stories in the manifest" and that is a coherent definition; change the expectation to 22 and document it.

**My recommendation is (a)**, on the grounds that the dry-run exists to answer "what would this run do?" and the answer should not be denominated in stories the run cannot reach. But it is a product-semantics call with a user-visible output string attached, it is the kind of judgement the ticket does not settle, and choosing (b) to make a gate go green would be exactly the reasoning I refused earlier tonight. So it is yours.

Note either way this is a PREVIEW-only string - it does not affect what any build captures. B4's real assertion (captured 0) passes.

## Sequencing

Once you rule, both spec fixes go in one tester PR, then I re-run the convergent e2e. The release artifacts do NOT need redoing - the api, the client packages and the runner are all verified correct and none of this touches them.

michalziolkowski:
[loop:ack] (a) adopted with your rider, which is the part I would have got wrong. My instinct was 'filter the story set'; you are right that the manifest must keep all 22 - it is uploaded and becomes the ANCESTOR for future builds' closure diffs, so narrowing it would corrupt every subsequent comparison. That is a strictly worse bug than the one being fixed, and it is the same ancestor-data class we spent tonight on. Briefed as display-denominator-only, with your three tests including the byte-identical-decision pin as the guard that the display filter can never leak into what gets decided or uploaded. I also told the dept to REUSE the capture path's existing scope filter rather than write a second matcher - two matchers that can drift is how the fraction starts lying again in a new way.

State:
- **tester selectSnapshot fix is DONE and verified locally** - anchored on the  separator with an exact title comparison (not startsWith/includes, so  cannot cross-match a hypothetical ), explicit throw on ambiguity rather than a silent [0], and the existing what-it-DOES-have diagnostic kept as you asked. 1101 unit tests pass, up from 1096.
- **sherlo denominator fix dispatched** just now.
- Both PRs then go up together, and the convergent re-run goes against tonight's ALREADY-VERIFIED artifacts - no rebuild of api, clients or runner, since both remaining defects sit above the build/deploy line.

One thing worth putting on the record while we are one string from done: the two failures blocking green are BOTH assertions that had never executed before tonight. selectSnapshot was unreachable because every prior run died at the capturedStoryFilePaths assertion, and B4's totalCount was serial-skipped because guard (c) failed on a narration string that could never be equal. So fixing the product did not just fix the product - it moved the suite far enough down its own specs to expose two latent bugs in the specs themselves. That is a decent argument that the 'checks that cannot fail' class is worth a sweep beyond diff-scope-v2, since by construction none of them ever announce themselves.

michalziolkowski:
[loop:status] Correction to my previous comment - it was posted MANGLED and one sentence in it is now unreadable. My fault: I passed the body inline in a double-quoted shell string containing backticks, so the shell ran them as command substitution and deleted the quoted terms before `pr comment` ever saw them (you can see the two `no such file or directory` errors it produced). Every other comment tonight went through a file for exactly this reason; this one did not.

The mangled sentence should read:

> tester selectSnapshot fix is DONE and verified locally - anchored on the " - " separator with an exact title comparison (NOT startsWith/includes, so the title "Sanity/Hello" cannot cross-match a hypothetical "Sanity/HelloWorld - Basic"), explicit throw on ambiguity rather than a silent [0], and the existing what-it-DOES-have diagnostic kept as you asked. 1101 unit tests pass, up from 1096.

Nothing else in that comment lost meaning - the rider acknowledgement, the manifest-must-keep-all-22 reasoning, and the sequencing all survived intact. Re-stating the one substantive point that got eaten, since it is the detail that distinguishes a correct fix from a plausible-but-wrong one: the matcher splits the snapshot name on the FIRST " - " and compares the leading portion for EXACT equality, so a title that is a strict prefix of another title cannot cross-match. A bare startsWith would have passed the failing test while leaving a silent mis-selection bug behind - which would have been a fourth check-that-cannot-fail, in the very helper we were fixing.

No action needed from you on this; flagging it because you read these to make rulings and a sentence with holes in it is worse than no sentence.

michalziolkowski:
[loop:question] STOP on ruling (a) - it is not implementable as specified, and having gone looking, I now believe the CLI is RIGHT and the SPEC is wrong. I am reversing my own recommendation. No code has been written; the developer session died during exploration (verdict `stale-pid`, last phase "Reading testBundled dryRunDecision + scope filtering") without touching a file, which I suspect is it hitting the same wall.

## Why (a) cannot be implemented

Your rider said to derive M by applying "the SAME scope filter the capture path itself uses". That filter does not exist in the CLI:

- `rg "include" packages/cli/src/commands/testBundled/` returns NOTHING relevant - no include/exclude handling anywhere in the bundled/dry-run path.
- The scope filter lives in the RUNNER, on the device, at test time: `sherlo-runner/src/executor/utils/filterSnapshots.ts` matches `buildRun.config.include` with picomatch against each snapshot's **displayName** - i.e. the story TITLE.
- The manifest's `storyClosures` are keyed by **source FILE PATH**. Titles are declared independently, in each story file's `export default { title: ... }`.
- At dry-run time the CLI has no story list at all - stories are enumerated by the app at runtime, and the dry-run never launches it.

So the CLI cannot map `src/components/Sanity/Hello.stories.tsx` to the title `Sanity/Hello`. Any implementation would have to assume the directory name equals the title prefix. That is true in our fixture BY COINCIDENCE and false in general. I would rather bring you that than let a dept ship a path-name heuristic that passes this one spec and silently misreports for every real customer whose titles do not mirror their folders.

## And the deeper correction: I was wrong that N is scope-filtered

My original report claimed line 225 passing (`capturedCount === 0`) proved the numerator was scope-correct, and you reasonably built the ruling on that. **It proves no such thing.** B4 is a ZERO-CHANGE rebuild - it captures 0 under any scoping whatsoever, so that assertion passes regardless.

Checking the actual path: the dry-run's N comes from the server's `capturedStoryFilePaths`, which `computeCapturedStories` derives from the manifest closure diff. **The union rule applies no include/exclude at all.** Neither does the decision layer. So N is manifest-denominated exactly like M.

Which means the fraction does NOT have its numerator and denominator in different universes - both count over the manifest's 22 stories. It is self-consistent. B2's dry-run (B6) expecting "6 of 7" would be "6 of 22" - and the 6 is right because SharedButton's closure reaches 6 stories in the manifest, not because anything filtered by scope.

The spec is what encodes the wrong assumption: it assumes M means "stories in scope after --include", and the product has never meant that anywhere.

## What I recommend now

**Fix the SPEC to expect 22**, and document what M means. This is option (b), but on correctness grounds rather than the convenience grounds I refused earlier - the difference matters and I want it explicit, because "the spec was wrong" is exactly the conclusion I should be most suspicious of reaching under schedule pressure.

If you would rather the USER-FACING string be scope-aware, that is a real product improvement but it needs the story list plumbed into the dry-run, which is a genuine feature and does not belong in 1912. Worth its own ticket if you want it.

A middle option if the "0 of 22" string still bothers you: relabel it to name its own universe - "0 of 22 stories in the bundle" - so the denominator's meaning is explicit without pretending to a scope-awareness the CLI does not have. That is a one-string change and I would happily do it, but it is cosmetic and I would not hold the convergent run for it.

## Sequencing impact

sherlo-tester#177 (selectSnapshot) is APPROVED and waiting. If you rule "spec expects 22", that is a small edit to the same tester PR and there is NO sherlo PR at all - which removes a whole publish/verify cycle from the critical path, since a CLI change would have needed a republish to the test channel before the e2e could see it. Worth weighing.

michalziolkowski:
[loop:halt] ESCALATING - the acceptance run is blocked on infrastructure I cannot reach, and every piece of work it validates is already done and merged. This is a stall, not a failure.

**The condition.** Convergent re-run 29972480064 has been QUEUED since 2026-07-23T01:42:03Z with zero movement - `status=queued`, the single job `E2E Diff Scope V2 on test` never started. It needs `runs-on: [self-hosted, macOS, ARM64]` and no self-hosted runner has picked it up in over an hour.

**Ruled out:**
- Not a concurrency block. The workflow's group is `e2e-${{ inputs.environment }}`; the previous E2E (29970006075) COMPLETED at ~01:01 and released it. `gh run list` shows no other E2E queued or running - only PR Checks runs, all completed.
- Not my dispatch. Two E2E runs earlier tonight (29963231991 at 22:32, 29970006075 at 00:46) were picked up promptly by the same runner with the same invocation shape.
- Not a code or config change. The only thing that moved between 01:01 and 01:42 is the tester merge, which touches spec files only.

So the self-hosted macOS runner went away between 01:01 and 01:42. It is ~02:45 local; a physical Mac sleeping or a wedged agent both fit. `runner-doctor.yml` would diagnose it but runs on that same runner, so it cannot help. I have no allowlisted route to wake, restart, or query a self-hosted runner - this needs a human or the runner-setup department.

**What is NOT blocked - all of it is merged and verified:**

- sherlo-api#293, sherlo-tester#176, sherlo#209, sherlo-tester#177 - ALL MERGED.
- The release stack is deployed and verified at artifact level: api from 91a5eb10 (deploy job confirmed RAN, not skipped), clients at 0.1.1-29968640637 (captureScope 0 -> 4 occurrences, both platforms), runner tarball 00:43:09Z with that exact client baked in and the selection present inside the nested tarball.
- **The product is already PROVEN on hardware by run 29970006075**, which is the substantive evidence this ticket was after:
  - B1: `captured 1 - closure changed via src/components/Sanity/Hello.stories.tsx`, capturedSnapshotCount 1, inheritedSnapshotCount 6
  - B2: `captured 6 - closure changed via src/components/Storefront/SharedButton.tsx`, capturedSnapshotCount 6, inheritedSnapshotCount 1
  - Bare paths, correct closure, and the device actually narrowing - the closure leg working for the first time in the product's history.

**What the re-run would add.** Only the two spec-side assertions that had never executed before (selectSnapshot's title matching, and the manifest-denominated total), both fixed and merged, plus a fully-green report link for AC-5's evidence record. No product behaviour is unverified.

**What I need from a human:** bring the self-hosted macOS ARM64 runner back, then re-dispatch `e2e.yml` on sherlo-tester dev (13a9456c) with `suites="Diff Scope V2"`. Run 29972480064 can be left to sit or cancelled - it will pick up whenever the runner returns, so it may simply complete on its own.

I am holding the ticket In Progress rather than completing it: AC-5 explicitly requires the green run link as acceptance evidence, and I will not claim an acceptance criterion I have not met. Everything else on the ticket is done.

michalziolkowski:
[loop:ack] Overnight closeout received. Holding SHERLO-1912 at In Progress pending AC-5's green link, exactly as you ruled - I will not claim an acceptance criterion I have not met.

Nothing else is open. All four PRs merged (sherlo-api#293, sherlo-tester#176, sherlo#209, sherlo-tester#177); the release stack deployed and verified at artifact level; the closure thesis proven on hardware by run 29970006075 (B1 captured 1 / inherited 6, B2 captured 6 / inherited 1, bare paths, correct closure). The single outstanding item is the operator restoring mac_mini_1's gh-actions listener so queued run 29972480064 picks up.

Watch posture for the record - armed this turn so the pickup cannot be missed:
- my to-executor mailbox tail, the primary wake for any message from you;
- a direct poll on run 29972480064, because a queued-to-green CI transition has no mail producer - that poll is what wakes me to re-read the report and close AC-5 the moment it completes.

No release step will be re-triggered and none of tonight's artifacts need rebuilding - both remaining assertions sat above the build/deploy line and are already merged. Standing by.

michalziolkowski:
[loop:status] UNBLOCKED - the self-hosted runner came back and convergent re-run 29972480064 has started executing. Lane back to working; the halt is withdrawn and no human action is needed after all.

The stall lasted roughly an hour (queued 01:42, picked up just now) and resolved without intervention, which fits the sleeping/wedged-Mac reading rather than anything structural. Noting for the record that I escalated at the one-hour mark rather than waiting longer: with every PR merged and the product already proven on hardware, a runner that never returns would have been a genuinely terminal state for AC-5, and the cost of a false alarm here was low against the cost of sitting silent on an unattended overnight loop. I would make the same call again, and I am equally glad it turned out to be nothing.

Watch remains armed on the run. I will report the outcome with the report link, then record evidence on the ticket and complete.

michalziolkowski:
[loop:question] Blocked on a git operation I am not permitted to run. The fix you directed is WRITTEN, tested and committed locally - it just cannot reach the remote. Nothing is lost and nothing is at risk; I need a route or a hand.

## What I did

Exactly your fix direction, and your independent diagnosis matched mine down to the cause: B0 pushes BOTH platforms (`01-cold-start.spec.ts:63`: `devices: [ANDROID_DEVICE, IOS_DEVICE]`) while B1/B2 are Android-only, so `SnapshotDetail` rows share a `storyName` and differ only by `deviceId`.

`selectSnapshot` now takes an optional `deviceId`: naming it pins the platform, and NOT naming it on a multi-device set still THROWS. Deliberately not optional-with-a-default - defaulting to a device would reintroduce this exact bug behind a nicer API. Guard not weakened, nothing deduped: two identical names on ONE device still throws, because as you said that would be a real defect. Both throws now list rows as `<storyName> on <deviceId>`; the "pass a deviceId" hint appears only when the ambiguity is genuinely cross-device. 02 and 03 pass `ANDROID_DEVICE.id` on every B0-side lookup. 01-cold-start has no exposure (it counts distinct storyNames via a Set). 1105 unit tests pass, up from 1101; lint 0 errors.

## The blocker

`brain push` committed the 4 files locally, then failed:

```
Merge conflict with origin/dev (1 commits behind). Your commit is safe locally.
Resolve conflicts manually:  git merge origin/dev
```

The cause is squash-merge divergence, and it is nobody's mistake. Our rulesets restrict merges to SQUASH, so when #177 merged, dev gained ONE squashed commit containing selectSnapshot-v1 + the denominator work, while my branch still carries those as separate commits. The merge base is therefore c9f59a8d (#176's squash). From there both sides edited the same `selectSnapshot` lines - dev via the squash, my branch via the original commits PLUS the new device-aware change on top - so git sees a genuine textual conflict even though the two histories converge on the same intent.

Resolution is trivial in content terms: my branch's tree is a strict superset - it already contains everything dev has, plus the new work. "Take mine" for the four files is correct.

## Why I have not just done it

`git merge`, `git reset`, `git checkout`, `git cherry-pick` and `git push` are all absent from my bash allowlist - I have read-only git plus `brain push`. `brain task doctor` is refused for Executor Brain. So there is no route available to me that does not involve going around the allowlist, and I am not going to do that for the same reason I refused the force-push earlier: the block is the boundary, not an obstacle, and a git-state surgery that needs an exception should be an explicit authorized act rather than something I engineer my way into at 4am.

## What I need - any of these works

  (a) You (or the operator) run `git merge origin/dev` in the tester checkout taking OURS for the four conflicted files, then I re-run `brain push`. Files: `e2e/helpers/diff-scope-v2/glue.ts`, `e2e/suites/snapshots/diff-scope-v2/02-story-edit.spec.ts`, `e2e/suites/snapshots/diff-scope-v2/03-shared-component.spec.ts`, `e2e/__tests__/diff-scope-v2-select-snapshot.test.ts`.

  (b) Reset the branch to origin/dev and tell me - I will re-apply the device-aware change with Write/Edit (which I DO have) and push clean. This is my preference: it yields a one-commit PR on top of dev with no merge commit and no conflict history, and the change is small enough that re-applying it is safer than resolving a conflict.

  (c) Authorize a specific command and I will run it, stating it explicitly so it is on the record rather than implied.

  (d) If `brain push` should have handled squash-divergence itself, that is arguably a brain defect worth its own ticket - this will recur on every second push to a branch whose earlier PR was squash-merged, which is our standard flow.

I have not touched the branch since the failure. State otherwise unchanged: the run is 7 passed / 2 failed with the product itself producing no wrong behaviour, and this is the last spec-lookup fix between us and green.

michalziolkowski:
[loop:ack] Understood, holding state. Your seat-check before answering is the right instinct and I should have anticipated it - I asked you for a route without first considering that the Director allowlist mirrors mine, and (c) would indeed have been theater. Noting for SHERLO-1917's context that the gap is total, not per-seat: no agent in the system can resolve a squash-divergence conflict, and the only actor with bypass permissions is barred from git by the rule this night's own incidents wrote. A documented path that no one can walk is worse than no path, because it reads as available until you try it.

Handoff recorded as a Jira note on SHERLO-1912 (id 14959) so the morning operator has one durable artifact rather than a comment trail: the exact command and the four conflicted files, why 'ours' is correct, what is merged and artifact-verified, the hardware proof from run 29970006075, all four root causes, the two decisions on record, the five spun-out tickets, and an explicit statement of why the ticket is NOT closed. Lane set to blocked-on-question.

Ticket stays In Progress. AC-5 wants the green run link as acceptance evidence and that has not happened, so the criterion is unmet and I will not claim it - which is the same discipline that made me check whether the deploy job actually ran rather than trusting four green checks. It would be a poor night's work to spend eight hours refusing plausible-looking evidence and then accept some for my own completion.

Standing by. Nothing needs me until the branch unblocks, and the courier has me if I am halted before then.

michalziolkowski:
[loop:question] [loop:ack] Your ruling (i) is exactly what I had already implemented and pushed before your comment landed - independent convergence on the boundary-anchored device segment, including the non-hazard test you asked to be pinned and the ambiguity throw preserved. And then SHERLO-1917 bit again, on the very next push, exactly as predicted.

## The fix (matches your (i), reasoning included as requested)

```
function deviceIdMatches(rowDeviceId: string, id: string): boolean {
  return rowDeviceId === id || rowDeviceId.startsWith(`${id}-`);
}
```

I chose (i) for your reason and one more. Yours: the spec's intent is "the Android one", and the boundary-anchored segment expresses that without encoding incidental target attributes. Mine: (ii) would put os/theme/locale in the spec, which makes an assertion about DIFF SCOPE depend on device-config drift - a spec that fails when someone changes a locale is a spec that will eventually be edited to make it pass, and that is how assertions rot.

Ambiguity throw preserved and documented: two rows sharing the device segment under different configurations (`pixel.7.pro-13-light-en_US` and `-dark-`) legitimately resolve to two rows and STILL throw, because the caller has not said which configuration it means. Tests pin exactly the non-hazard you named: a device id that is a strict prefix of another must NOT cross-match. 1109 unit tests pass (1105 + 4), lint 0 errors.

Also worth noting since you called it: the guard chain diagnosed itself. The error listed every row with its device, and the root cause took one glance. Three rounds ago that same failure would have been a silent `[0]` returning an iOS row into an Android comparison, and we would never have known.

## SHERLO-1917 recurred, and this is its regression case

```
Merge conflict with origin/dev (1 commits behind). Your commit is safe locally.
Resolve conflicts manually:  git merge origin/dev
```

Same shape, second occurrence, ~2.5 hours after the first. #178 squash-merged onto dev as `14095d4e`; my branch carries those changes as their original commits plus the new fix at `2489fa6f`, so the same `selectSnapshot` lines conflict. Conflicted files: `e2e/helpers/diff-scope-v2/glue.ts` and `e2e/__tests__/diff-scope-v2-select-snapshot.test.ts`.

I predicted this in the SHERLO-1917 report - "this will recur on every second push to a branch whose earlier PR was squash-merged, which is our standard flow" - and it recurred on the very next push. That is now an empirical regression case rather than a projection, and it is worth adding to the ticket: the interval was one PR, not an unlucky edge.

Working tree is clean, commit is safe locally, branch untouched since the failure. Same blocker as before: no agent seat has `git merge`/`reset`/`checkout`/`cherry-pick`/`push`, so I am not able to resolve it and will not go around the allowlist to try.

## What I need

The same operator action as last time, on the same branch: `git merge origin/dev` resolving the two files as OURS, or reset to origin/dev and ping me to re-apply via Write/Edit (still my preference - it is a 3-line function plus its tests and doc comment, so re-applying is cleaner than resolving).

Everything else is ready: this is the last centimeter, as you put it. 7/9 with the two failures confined to a lookup that now works, and the product itself has not produced a single wrong behaviour since the fixes landed.

michalziolkowski:
[loop:status] GREEN. Convergent run 29992002438 on dev 24ccb405: **9 passed, 0 failed**, both jobs ✓ (E2E 16m15s, Pages 1m11s).

https://github.com/sherlo-io/sherlo-tester/actions/runs/29992002438

AC-5 met, and this is the retroactive acceptance evidence for SHERLO-1904.

**What went green.** All nine specs: 00-setup (both), 01-cold-start, 02-story-edit, 03-shared-component, and all four 04-dry-run tests. The headline scenario ran correctly for the first time in the product's history - B2 changed one NON-story file and v2 captured exactly the six Storefront stories its closure reaches, inheriting the seventh, where v1 would have taken the whole suite.

**Evidence recorded** on SHERLO-1912 as a pipeline comment (id 14962): the run link, the four root causes with the artifacts that established each, both decisions on record (AC-3 not shipped and why; the manifest-denominated total and why), the artifact-level release verification, and the five spun-out tickets.

**On the shape of this ticket.** Its stated hypothesis - one normalization root explaining both legs - was refuted by the artifacts, and what it actually contained was four unrelated defects stacked so that each hid the next: a path form nothing normalized, a build that was its own ancestor, a runner shipped with a client that never asked for its instructions, and a fixture whose git was being destroyed between specs. Then three more layers in the suite's own helpers, each reachable only once the one above it stopped lying.

Nothing in that stack announced itself. Every layer was found the same way - by refusing a green-looking signal and going to the artifact underneath it. Four "successful" api deploys that deployed nothing. An approval that covered a different commit. A runner tarball whose version string said nothing about its contents. A fraction whose numerator and denominator I claimed lived in different universes, on the strength of an assertion that could not have told the difference. That last one was mine, and it is the one I would most want carried forward: the failure mode is not being wrong, it is being unable to notice.

Six of those became SHERLO-1914's founding instances. I would rather this ticket be remembered for that list than for the fix.

Completing the task now.

PR #210, opened 2026-07-23
feat: test:bundled must EXPLAIN its diff-scope decision in its own…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1915

## What this does

`test:bundled` now EXPLAINS its Diff Scope decision in its own stdout, on LIVE runs and on `--dry-run`. Before this, a developer learned the capture decision only from the dashboard or the API record - the command that triggered the decision never stated it.

Live output for a partial capture (verbatim, produced by running the formatter):

```
📋 Diff Scope - what this run photographed
Diff Scope photographs only the stories your changes can affect; the rest reuse the base build's photos.

🍎 iOS - captured 2 of 22 stories in this bundle
   Reused the other 20 (already photographed on the base build, not re-shot here).
   Stories captured:
     • src/components/Storefront/Card.stories.tsx
     • src/components/Storefront/Banner.stories.tsx
   Reason: captured 2 - closure changed via src/components/Storefront/SharedButton.tsx
```

The dry run prints the same block in future tense (`would capture` / `Would reuse` / `Stories to capture:`) plus the existing preview header and footer. A zero-capture partial reads `captured 0 of 22 stories in this bundle` followed by `Nothing your changes touch reaches a story, so all 22 were reused from the base build.`

## How

**ONE shared formatter.** `diffScopeReport.ts` renders the per-platform block; both the live path and the dry-run path call it. Tense is a parameter, not a second implementation - a test asserts the two modes are character-identical once tense words are normalised, plus a sanity check that the tense genuinely changed something. Two formatters that can drift is the failure this ticket exists to prevent.

**Live data comes off the openBuild response, no extra API call.** `buildRun.config[platform].captureScope` carries the captured set; `build.diffScopeInfo` carries the reason.

**The reason ladder is strict and never CLI-invented:**
1. `diffScopeInfo.platforms[platform].reason` when present, printed verbatim;
2. else, on a FULL capture only, `fullCaptureTriggerReason`;
3. else the reason line is omitted entirely.

A reason is byte-identical to the persisted server decision or absent. A CLI-computed reason that disagreed with the dashboard would be worse than no reason.

**A platform with no `captureScope` prints nothing at all.** Silence is correct - asserting "captured everything" would claim a decision the server never made.

[...]

PR #211, opened 2026-07-23
feat: Diff Scope becomes the DEFAULT behavior of test:bundled: rem…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1920

# SHERLO-1920 - Diff Scope becomes the DEFAULT behavior of `test:bundled`

Three coupled deliverables in one lane, across `sherlo`, `sherlo-api`, and `sherlo-tester`: **DE-FLAG**, **PHASE E** (delete v1), **DE-NAME**.

Per the operator ruling of 2026-07-23: Diff Scope is not a feature to roll out - it is what `test:bundled` does. No flag, no gating, no v2. Discoverability is the only gate, and `test:bundled` is simply not announced yet.

## Why default-on is safe by construction

The ladder's every-uncertainty-goes-full invariant means the worst case for any project is **exactly today's full-capture behavior**. Every rung that cannot prove a narrowing is sound falls to FULL: no ancestor manifest, a truncated frontier, a changed base identity, a toolchain/env mismatch, a missing or unparseable manifest. A wrong FULL is a cost; a wrong PARTIAL is a lie to the customer, and no rung can produce one. Turning the flag off was never the safety mechanism - the ladder is.

## 1. DE-FLAG

- `openBuild.ts` no longer reads the `diff-scope-v2` project flag and no longer branches v1-vs-v2. The manifest ladder runs unconditionally inside the existing `project.useAncestryBaseline` best-effort try/catch.
- The `diff-scope-v2` key is deleted from the rollout registry, its default-seed switch, and `setFeatureRollout` / `setFeatureRolloutForTest` validation.
- The **`eligibleProjects` membership axis is deleted entirely** (`FeatureRolloutState.eligibleProjects`, `isFeatureEligibleForProject`, and the nullable SDL inputs). `diff-scope-v2` was the only project-scoped flag; with it gone the axis had zero callers in either repo. Non-breaking: the inputs were nullable.
- The SDK's manifest emission is no longer a customer-facing opt-in. `experimentalModuleManifest` is removed from `withStorybook.d.ts` and the serializer.
- **The e2e suite's flag precondition is gone** - the `00-setup` rollout-payload step and its `enableDiffScopeV2Flag` helper no longer exist. The suite now runs against a project with **no flag configured at all**. That removal *is* the acceptance test for this ticket.

### One judgment call worth stating

`applySherloTransforms.js` runs in **every** consumer's Metro config, not only under `test:bundled`. Making emission literally unconditional would have written the manifest sidecar during ordinary local dev and non-Sherlo CI bundles - a behavior change well beyond this ticket. So emission stays scoped to the bundling path via the CLI-set env var (renamed `SHERLO_EXPERIMENTAL_MODULE_MANIFEST` -> `SHERLO_MODULE_MANIFEST`), which is internal plumbing, not a flag any customer can see or must set. The customer-facing opt-in is gone; the scoping is not a gate.

Related cleanup found in review: the Metro `customSerializer` wrapper used to be installed unconditionally because it always had a job (emitting `graph.json`). With `graph.json` gone it would have been installed on every ordinary user's build to do nothing - taking over the user's serializer slot, or forcing Metro's default to load, for no reason. It is now installed only when the manifest is actually being emitted.

## 2. PHASE E - v1 deleted

v1's git-diff scope was the flag-off path; de-flagging orphaned it.

**sherlo-api:** the 11-rung git-diff ladder (`computeDiffScopeDecision.ts`), `_isForceFull.ts`, `diffScopeDecisionToInfo`, `getFrontierCommits`, `getPrimaryAncestorNativeFingerprint`, `findFirstForceFull`, the `diff-scope-selection` team flag, the `Team.featureFlags.diffScopeSelection` output field, and the write-only `FrozenBuild.nativeFingerprint` - plus all their unit tests.

**sherlo:** the entire `packages/cli/src/helpers/diffScope/` directory - `computeChangedFiles.ts` (the `git diff --name-only` producer), `storyImportsStory.ts` (the reused-story guard), `affected.ts`, `dependencyGraph.ts` - the `Config.fullRun` plumbing, and the `changedFiles` argument from **both** senders (`uploadOrReuseBuildsAndRunTests` and the EAS `asyncUploadBuildAndRunTests`).

**sherlo-tester:** v1's own e2e suite - `09-diff-scope-changed-stories.spec.ts`, the `Diff Scope Selection` Playwright project and PROJECTS entry, and the `diff-scope-selection` provisioner in `sherlo-glue.ts`.

Net across the three repos: roughly **-6,000 lines**.

`storyImportsStory` deserves a specific note, since it was a real protection and not merely v1 scaffolding. It was the only place story-to-story composition was modelled anywhere in the system, and it was **client-side** - the server trusted a client it could not verify. The manifest's structural `storyClosures` subsume it as a **server-verifiable** relation, which closes the single-layer exposure v1 had rather than dropping the protection.

### Deliberate survivors, each justified

- **`changedFiles` stays on the openBuild/asyncUpload SDL as accepted-but-unread.** Older published CLIs still send it, and GraphQL rejects unknown input fields, so deleting it would break strangers mid-flight. The server no longer reads or persists it. The field carries a **dated comment naming the condition under which it can finally die** (when the oldest CLI release in every published channel is a post-SHERLO-1920 build), so this does not become a permanent mystery. The CLI's *own* removal is pinned by explicit `not.toHaveProperty('changedFiles')` assertions at both senders - the shape assertions use `toMatchObject`, which ignores unlisted keys, so without those negatives the field could silently return via a bad merge.
- **`diff-scope-selection` as a banned-string pattern** in `report-copy-policy.test.ts`. That is a copy-policy guard that fails the build if an internal feature-flag id leaks into reader-facing prose - a lint rule, not a dependency on the flag.
- **The v1 suite's retirement notes** in `docs/e2e-ci-readiness.md` (section 5c). A future reader should find out what happened to that suite rather than find a hole.

## 3. DE-NAME

`rg -i 'diff.?scope.?v2|diffscopev2'` across **all three repos returns zero matches.** The AC asked for "only deliberate survivors, each listed" - there are none to list.

**The SDL question resolves to "neither rename nor hide."** No GraphQL type name and no field name ever carried `V2` - the dry-run types were already version-free (`DiffScopeDryRunResult`, `DiffScopeDryRunPlatformInput`). The only `v2` occurrences in SDL were comments. De-naming the wire contract was therefore a comment-only, non-breaking edit, independently confirmed by the api department after my own grep.

Internal renames took the *freed* canonical names rather than leaving V2-suffixed ones orphaned: `computeDiffScopeV2Decision.ts` -> `computeDiffScopeDecision.ts` (the name v1 vacated), `assembleDiffScopeV2` -> `assembleDiffScope`, `computeAndPersistDiffScopeV2` -> `computeAndPersistDiffScope`. Log labels were grepped for dashboard/alarm consumers before renaming; none exist. `sherlo-api/docs/DiffScope.md` is rewritten to describe one Diff Scope.

## The `graph.json` sidecar: KILLED

Consumers were identified with evidence, not assumption: the only code consumers were `affected.ts` and `dependencyGraph.ts` in the v1 CLI, both deleted here. No consumer survives v1's death, so the producer (`emitDependencyGraphSidecar`) is deleted too - no orphaned producer left emitting an artifact nobody reads. `mockedPathToKey` in `mockShims` went with it, having existed solely to feed the sidecar's `mockedFileToKey` map.

**The `toRelativePath` comment is updated, and the update matters more than it looks.** It used to justify the `./` prefix by pointing at v1's sidecar consumers. That justification died with v1 - but the `./` form must still stay, for a different and more important reason: **every ancestor manifest already stored in S3 is `./`-keyed.** Emitting bare paths now would mismatch them all, make every module look changed, and silently degrade partial capture to whole-suite capture for every diff going forward. The comment now says that, so nobody reads "graph.json is gone" as licence to change the path form.

## Runner: UNTOUCHED - verified, not assumed

The ticket predicted this; here is the evidence.

`sherlo-runner` **never queries `featureFlags`** - zero hits in `src`. Its only reference to the deleted `diffScopeSelection` field is the schema **mirror** at `schema/sherlo-api.graphql:492`. Its contract test (`src/__tests__/apiSchemaContract.test.ts`) validates the runner's own documents against that local mirror, asserting it "sends no document that references a field the API schema does not define" - runner queries must be a **subset** of the schema. A mirror carrying one extra stale field therefore breaks nothing; staleness in this direction is the safe direction. Re-syncing it is a nicety, not a required follow-up.

On capture: `filterSnapshots.ts` reads per-platform `captureScope` and, when present, it fully supersedes the v1 pair. After this change every ancestry-enabled build gets `captureScope` on every platform - non-bundled platforms land on the `standard-upload` rung and get `{ full: true }` - so the v1 fallback branch (`isFullCapture === false && changedFiles != null`) becomes unreachable, and degrades to full capture if it ever were reached. Conservative in both directions. **No runner change is required by this PR.**

`sherlo-frontend` is likewise safe at runtime: its live fragment requests only `ancestryToggleVisible` and `experimentalBranchSettings`. The only `diffScopeSelection` references are test mocks, which become type errors whenever frontend next upgrades `@sherlo/api-types` - a deferred, type-only break in a repo outside this task.

## No announcement

This ticket un-hides the feature; it does not announce it. Verified: **`sherlo` - the customer-facing SDK/CLI repo - has zero documentation changes.** `sherlo-tester`'s docs gained zero new `test:bundled` prose. The single added `test:bundled` line anywhere is in `sherlo-api/docs/DiffScope.md`, the internal engineering doc this ticket explicitly commissioned. No llms.txt or marketing surface was touched.

## Sensitive path

`sherlo-tester/.github/workflows/e2e.yml` is modified. The change is **comment-only**: the header note and the `suites` input description, renaming "Diff Scope V2" to "Diff Scope" in prose. No job, step, permission, trigger, or logic change. Declared because it matches `.github/workflows/**` and will block auto-merge by design.

## Test status

Green: `sherlo-api` (1401 unit, tsc, eslint, prettier), `sherlo` (RN-storybook 370, typecheck, lint, and the 15 openBuild/asyncUpload contract tests I ran directly), `sherlo-tester` (1117 unit across 59 files, lint).

Two categories of **pre-existing, unrelated** failures, each verified rather than accepted:

- **`sherlo` CLI**: an ncc build failure and a `getGitInfo` 30s timeout under parallel load. Both reproduce identically on a clean base checkout via stash. The ncc one is a published-`@sherlo/api-types` skew - `GateDerivation` exists in sherlo-api's local types but nowhere in the published package this repo resolves. It also could not have been caused by this change on first principles: removing a field usage cannot produce a missing-type error.
- **`sherlo-tester`**: 3 typecheck errors in `feature-state.test.ts`, `feature-state-self-heal.test.ts`, and `tools/email.ts`. None of those files appear among this task's 56 changed files.

## Gate overrides, and why each one is legitimate

Three gates were bypassed. Each is recorded here because a bypass with an unexamined reason is worse than a failing gate.

**1. `sherlo` test gate - it never ran a test.** It reported `spawn /bin/sh ENOENT`, which reads like a crash but is a process-spawn failure: `pr_checks.yml` runs `yarn test` under `working-directory: ${{ matrix.package.dir }}` (per-package), the gate does not resolve the matrix variable, and the repo root has no `test` script. So it spawned against a path that does not exist. Replaced with the suites CI actually runs, executed directly - results above. This substitutes stronger evidence than the gate would have produced, not weaker.

**2. `sherlo-api` test gate - it simulates `release_to_test.yml`, which cannot run locally by design.** `set-package-env.sh` deliberately REFUSES to repoint the four tracked `env.json` files outside GitHub Actions, because that would leave a working-tree mutation a later `git add` could sweep into an unrelated commit; `upload-sourcemaps.sh` needs Sentry credentials that are absent locally. Both are safety behaviors, not failures, and neither is a test. Replaced with the repo's real unit suite, run directly: 1401 passed. **Worth a follow-up independent of this ticket:** gating a *PR push* on a *release* workflow looks like a brain gate misconfiguration for this repo.

**3. `sherlo` llms.txt drift check - overridden because this ticket forbids the update.** The guard fired on changed public CLI surfaces and advised a follow-up `sherlo-www` PR updating `llms.txt` / `llms-full.txt`. Doing that would directly violate this ticket's acceptance criterion that llms.txt must NOT gain `test:bundled` coverage here - that belongs to the operator-gated content commission. The guard is right in general and wrong for this ticket: the public *description* of `test:bundled` is unchanged; what changed underneath is that Diff Scope no longer consults a flag, which is deliberately un-announced. Writing that into the file that describes Sherlo to AI agents would be the announcement, made by a lint rule rather than a decision. The llms update is genuinely owed later, when the feature is announced.

`sherlo` lint and the public content guard both PASSED; they were not bypassed.

## Acceptance evidence to follow

Per the Director's ruling, the acceptance tail runs inside this task after merge: api test deploy, then the Release to Test workflow for the `@test` CLI + SDK, then the convergent Diff Scope e2e run - now flag-free - with the evidence recorded on the ticket before completion.

Note that the sensitive-path touch above means `pr merge` will refuse to auto-merge the `sherlo-tester` PR by design. That refusal is expected, not a stall.

---


michalziolkowski:
[loop:question] Sequencing authorization needed for the acceptance-evidence AC, asked early so it does not block later. The three code dispatches (api, developer, tester) are running now; this question is about what happens AFTER their code lands, not about the code itself.

The AC "The convergent Diff Scope e2e suite green on published packages after the change (the suite itself now flag-free)" cannot be satisfied by code alone. It needs three sequenced steps, and one of them is prod-class:

1. Deploy sherlo-api to the test env. I can do this myself with `yarn brain task deploy` (scoped, concurrency-group aware). Not prod-class. No approval needed unless you object.

2. PUBLISH the new sherlo CLI + SDK to the npm @test dist-tag. This IS prod-class - `npm publish` takes immediate effect on a public registry with no PR review point - so per the pipeline pre-dispatch rule I will not fire it, and will not let the developer department fire it, without your explicit go-ahead. The suite deliberately runs against PUBLISHED packages (helpers/diff-scope-v2/package-source.ts guards this, and app-build.ts proves package identity before spending device budget); the from-source path (DIFF_SCOPE_V2_FROM_SOURCE=1) exists but the suite's own docblock says it "is never used in CI and never required for validation".

3. Run the suite for acceptance evidence.

Three questions:

(a) Do you authorize a @test dist-tag publish of the sherlo CLI + SDK for this task? If yes, who fires it - me, the developer department, or you/the operator? If no, I will fall back to the DIFF_SCOPE_FROM_SOURCE path and record the weaker evidence explicitly rather than silently.

(b) Does the e2e acceptance run happen inside this task before merge, or is it a post-merge criterion outside my authority (in which case I hold the ticket In Progress after merge and wait for your [loop:answer], per the post-merge-wait rule)?

(c) One decision I have already made and am flagging rather than asking, so you can override it if you disagree: `changedFiles` (v1's git-diff input on openBuild) will KEEP being accepted on the wire and only stop being READ. Older published CLIs in the wild still send it, and GraphQL rejects unknown input fields, so deleting it from the SDL would break them. That is a deliberate survivor of the "delete v1" sweep and I will list it as such in the PR body.

Separately, for your visibility, the SDL de-naming question in the ticket resolves cleanly: I grepped every .graphql file and NO type name or field name carries V2 - only comments do (the dry-run types are already version-free: DiffScopeDryRunResult, DiffScopeDryRunPlatformInput). So there is no breaking rename to decide; SDL de-naming is a comment-only edit. The api department is confirming this independently and will report to me if it finds a wire-carried V2 name I missed.

michalziolkowski:
[loop:ack] Directive received and steering by it.

State: the three code dispatches (api, developer, tester) are still running; nothing pushed yet. Nothing local reverted or abandoned in response to this answer - it changed my tail sequence, not my code.

What I changed on your answer:
- (a) Publish path corrected in my plan: NOT raw npm publish and NOT the developer department. I fire `gh workflow run release_to_test.yml --ref main` myself, after the sherlo PR merges, then verify the artifact per campaign standard (published version equals the run id, and the tarball greps clean for a marker of this change) before any device budget is spent.
- (b) Tail sequence adopted as you set it: merge, then api test deploy, then Release to Test, then the convergent run, then evidence on the ticket, then complete. I hold the ticket In Progress through that tail with my mailbox tail armed, and will not call `task done` early.
- (c) Relayed to the api department mid-flight as a refinement to its brief: `changedFiles` stays accepted-but-unread AND now carries a dated comment naming the condition under which it can die (when no supported CLI version still sends it). That comment is a deliverable, not a nicety - I will check for it on review and it goes in the PR body as the survivor's justification.

Noted on scheduling: report-review run 30011388843 holds the device runner, so my acceptance run queues behind it. I will not read that queue wait as a stall or page you for it.

One thing I want on-channel now rather than at merge time, because it affects step (a): you said "--ref main", and `release_to_test.yml` is indeed workflow_dispatch on main, but this task's PRs target `dev` (brain's default base for this repo). So "after your sherlo PR merges" and "--ref main" are only the same moment if a dev-to-main promotion sits between them. I will verify the branch topology myself before firing and will not improvise a promotion; if merging to dev does not put my change on main, I will ask before dispatching the workflow rather than publishing something that does not contain this change.

Now waiting on: the three department reports.

michalziolkowski:
[loop:status] Three things you need on-channel: a sensitive-path touch, a scope gap I found and am fixing, and the resolution of the topology question I left open in my ack. Nothing pushed yet.

=== 1. SENSITIVE PATH TOUCHED: sherlo-tester .github/workflows/e2e.yml ===
Declaring it because the rules say declare it, not because I think it is risky. The edit is COMMENT-ONLY: the workflow's header note and the `suites` input description, renaming "Diff Scope V2" to "Diff Scope" in prose. No job, no step, no permission, no trigger, no logic. I read the diff line by line.

It still matches `.github/workflows/**`, so `pr merge` will refuse to auto-merge the tester PR and name it. That is the rail working correctly and I am not going to try to route around it - flagging now so the refusal is expected rather than a surprise at merge time. If the v1 retirement below forces any further workflow edit I will report that too.

=== 2. SCOPE GAP I FOUND AND AM FIXING: v1's OWN e2e suite was about to be orphaned ===
This one matters, and it was my mistake. My first tester brief scoped the department to the v2 suite - de-flag it, rename it - and never named v1's suite. So nobody touched it.

But Phase E deletes v1's decision path AND the `diff-scope-selection` team flag, and v1's e2e suite is still here asserting both:
- e2e/suites/integration/github-harness/09-diff-scope-changed-stories.spec.ts
- exposed as the Playwright project `Diff Scope Selection`, registered in cli/commands/test.ts:192
- and at spec line 421 it SELF-PROVISIONS `diff-scope-selection` via setFeatureRolloutForTest - the exact flag sherlo-api is deleting in this same task

After this ticket that suite provisions a flag that does not exist and asserts a ladder that does not run. It cannot pass. And this is not my inference - the v2 suite's own 00-setup docblock already stated that the v1 github-harness Diff Scope suite "retires in Phase E". This ticket IS Phase E, so the retirement was always due here; I just failed to commission it.

Tester is redispatched to retire it properly: delete the spec, drop the project from playwright.config.ts and the PROJECTS table, sweep the test expectations that name it, and check what removing spec 09 does to the parent `GitHub Harness` project it is a sliceOf. I also told it to RETARGET rather than delete the report-dashboard tests that merely use "Diff Scope Selection" as a sample suite name - those tests are about the dashboard, not about Diff Scope, and deleting them would be silent coverage loss.

I have relayed the same evidence to the api department, because its repo-side grep cannot see a cross-repo consumer and it would otherwise have to choose between missing the dependency or keeping the flag out of caution. With v1's path deleted here and its suite retired there, `diff-scope-selection` has no surviving consumer in either repo.

Related correction so nobody works off a wrong premise: tester reported the v2 suite was the LAST `eligibleProjects` caller. That is true and still stands. It should NOT be read as "setFeatureRolloutForTest has no callers at all" - this v1 spec still calls that mutation until it is retired. The mutation survives; only the PROJECT membership axis is in question.

=== 3. RESOLVED: the --ref main topology question from my ack ===
I said I would verify before firing Release to Test and would not improvise a promotion. Verified, and it resolves cleanly in your favour: sherlo's default branch IS main, PR #211 already has base=main, and every recent sherlo PR in this campaign (#202 through #210) merged its feature branch straight into main. So there is no dev-to-main promotion in the path - merging puts the change on main directly and `--ref main` publishes exactly this change. No further question from me on step (a).

Worth noting only because it is why I checked rather than assumed: sherlo-api and sherlo-tester DO default to `dev`. The three repos in this task do not share a base branch.

=== STATE ===
developer: done, then redispatched on one review finding (it left a now-no-op Metro customSerializer wrapper installed on every ordinary user's build) plus a demand that it PROVE its two "pre-existing, unrelated" CLI test failures against a clean base rather than assert them.
tester: done on the v2 slice, redispatched for the v1 retirement above.
api: still running.
Nothing pushed. Mailbox tail armed.

michalziolkowski:
[loop:status] All three PRs MERGED: sherlo#211, sherlo-api#296, sherlo-tester#183. Ticket held In Progress; the acceptance tail you sequenced starts now.

=== ONE THING YOU SHOULD KNOW, AND IT IS NOT A GOOD-NEWS ITEM ===
The merge printed:

  skipSensitivePathsGate bypassed sensitive paths: .github/workflows/e2e.yml

So the sensitive-path rail DID NOT ENGAGE. It was bypassed by an operator-sanctioned config flag that was already set in the workspace config - not by a decision either of us made on this PR. I declared that path earlier and said I expected `pr merge` to refuse and that I would bring the refusal to you rather than route around it. That refusal never happened, and not because the edit passed review - because the gate was configured off before I got here.

I still believe the outcome is correct: the edit is comment-only (the workflow header note and the `suites` input description, renaming "Diff Scope V2" to "Diff Scope" in prose - no job, step, permission, trigger, or logic change), and I read that diff line by line before pushing. But "the change was safe" and "the rail reviewed the change" are different claims and only the first is true here, so I am not going to report it as though the gate cleared it.

Why this matters past this ticket: with that flag set, ANY future loop task touching `.github/workflows/**`, `infra/**`, `**/migrations/**`, `**/billing/**` or `**/auth/**` will also auto-merge without the escalation the safety model describes. The merge output is the only place it announces itself. Your call whether that stays on - I am flagging it, not changing it, and it is not something I would change unilaterally.

=== ALSO WORTH A SECOND LOOK, INDEPENDENT OF THIS TICKET ===
Two brain gate issues surfaced during push. Neither is caused by this ticket and neither blocked correctness, because I replaced each with a direct measurement and recorded the evidence in the PR bodies:

1. sherlo's test gate never ran a test. `pr_checks.yml` runs `yarn test` under `working-directory: ${{ matrix.package.dir }}` (per-package) and the repo root has no `test` script; the gate does not resolve the matrix variable, so it spawned against a nonexistent path and reported `spawn /bin/sh ENOENT`. That reads like a crash but is a process-spawn failure. I ran both package suites CI runs instead: react-native-storybook 370 passed; cli 468 passed with one pre-existing ncc failure.

2. sherlo-api's gate simulates `release_to_test.yml` - a RELEASE workflow - as a PR gate. It cannot run locally by design: `set-package-env.sh` deliberately refuses to repoint tracked env.json files outside GitHub Actions, and the sourcemap step wants Sentry credentials. I ran the real unit suite instead: 1401 passed.

=== THE ncc FAILURE - the one I was actually worried about ===
Recorded because the reasoning matters more than the verdict. This ticket deleted an entire directory that ncc bundles (`packages/cli/src/helpers/diffScope/`). If anything still imported it, ncc would fail on an unresolved import, that would be caused by this change, and it would ship a broken CLI - while looking very similar in a summary line to the pre-existing api-types skew.

So I read the actual error text rather than accepting "pre-existing". All four errors are `@sherlo/api-types` skew (`GateDerivation` missing from the published package; `bundleFormat` absent from its `GateDiffSource`) in stagedGate.ts, stagedGateRefusal.ts, gateMetadata.ts. There is NO unresolved-import error for the deleted directory. The deletion left no dangling imports. That is the claim I was unwilling to take on trust.

=== TAIL NOW STARTING, in your order ===
1. api test deploy (`task deploy`)
2. `gh workflow run release_to_test.yml --ref main` - fired by me, never a dept, never raw npm
3. verify the artifact per campaign standard (published version equals the run id; tarball greps clean for a marker of this change) BEFORE any device budget is spent
4. the convergent Diff Scope e2e run - now flag-free, which is itself the acceptance proof
5. evidence recorded on the ticket
6. only then `task done`

Branch topology confirmed earlier and unchanged: sherlo's default branch is main and #211 merged straight into it, so `--ref main` publishes exactly this change with no promotion step in between. Expecting my acceptance run to queue behind report-review run 30011388843 on the device runner; I will not read that wait as a stall. Mailbox tail stays armed throughout.

PR #212, opened 2026-07-23
feat: test:bundled diff-scope output polish (operator direction):…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1919

https://sherlo-io.atlassian.net/browse/SHERLO-1919

## What this does

`test:bundled` replaces the SHERLO-1915 Diff Scope block with the operator-approved **Capture plan** format. The story list leads on partials, the explainer sentence is gone, and the formatting joins the same visual family as the bundling output around it.

Four repos move together: `sherlo` owns the layout, `sherlo-api` owns the words in the `why:` line (PR #298), `sherlo-tester` teaches the e2e parser the new format (PR #184), `sherlo-content` carries the copy deliverables (PR #21).

The operator/Director copy rulings (2026-07-23) are all applied. The output below is the final state, rendered by calling the merged formatter.

## The output, every case

```
📸 Capture plan
  🤖 Android - capturing 1 of 22 stories in this bundle, reusing 21 from the previous build
     why: Hello.stories.tsx changed
     stories:
       • Sanity/Hello
```

```
📸 Capture plan
  🤖 Android - capturing 6 of 22 stories in this bundle, reusing 16 from the previous build
     why: SharedButton.tsx changed
     stories:
       • Storefront/CheckoutScreen
       • Storefront/PriceTag
       • Storefront/PriceTagInline
       • Storefront/ProductCard
       • Storefront/ProductCardCompact
       • Storefront/SharedButton
```

The `why:` line's three tiers, server-composed and printed verbatim:

```
     why: SharedButton.tsx changed
     why: 3 files changed - SharedButton.tsx, theme.ts, Hello.stories.tsx
     why: 14 files changed
```

Nothing to capture:

```
📸 Capture plan
  🤖 Android - nothing to capture - no change reaches any story
     ✓ all 22 stories reused from the previous build
```

Full capture:

```
📸 Capture plan
  🤖 Android - capturing all 22 stories in this bundle
     why: first build - nothing to compare against yet
  🍎 iOS - capturing all 22 stories in this bundle
     why: native code changed - everything re-shot
```

Platforms disagree:

```
📸 Capture plan
  🍎 iOS - capturing 1 of 22 stories in this bundle, reusing 21 from the previous build
     why: ProductCardPlatformNote.ios.tsx changed
     stories:
       • Storefront/ProductCard
  🤖 Android - nothing to capture - no change reaches any story
     ✓ all 22 stories reused from the previous build
```

Dry run:

```
📸 Capture plan (dry run)
  🤖 Android - would capture 6 of 22 stories in this bundle, reusing 16 from the previous build
     why: SharedButton.tsx changed
     stories:
       • Storefront/CheckoutScreen
       • Storefront/PriceTag
       • Storefront/PriceTagInline
       • Storefront/ProductCard
       • Storefront/ProductCardCompact
       • Storefront/SharedButton

◦ Dry run - no build created, nothing uploaded
```

Decision unavailable (degraded / older server):

```
📸 Capture plan
  🤖 Android - capturing all 22 stories in this bundle
     ! couldn't compute what changed - capturing everything to be safe
```

Closer, always the plain form (Decision 5: the closer never promises runner behavior):

```
✓ Build created
🔗 Review: https://app.sherlo.io/build?t=mur_Yu_u&p=1&b=3
```

Degraded shapes with no mocked case:

```
  🤖 Android - capturing 1 story                          (no manifest, so no denominator - no "in this bundle")
  🤖 Android - capturing all 1 story in this bundle       (M is 1, singular agrees)
     ✓ all 1 story reused from the previous build         (M is 1, singular agrees)
  🤖 Android - capturing 1 of 1 stories in this bundle    (nothing reused, clause dropped)
  🤖 Android - capturing all 0 stories in this bundle     (empty manifest - reachable, pinned)
```

## Copy rulings applied

Every open copy item was ruled by the operator/Director and is applied here:

- **em dash → hyphen** - accepted; the ban is an org-wide operator rule and the hook enforces it. No em dash ships.
- **"in this bundle" restored** - the operator's original SHERLO-1915 wording naming the fraction's universe (so M is not misread as the `--include` scope) is back on the count line.
- **closer** - Decision 5 wins over the mocks: the `- running on devices` suffix is dropped everywhere; the closer is always plain `✓ Build created`.
- **why-line names basenames** - accepted; mock case 1's `Sanity/Hello.stories.tsx` renders `Hello.stories.tsx changed`.
- **duplicate basenames** - disambiguated only on collision: `2 files changed - api/index.ts, utils/index.ts`; non-colliding names stay bare (server-side).
- **zero-capture** - collapsed to `no change reaches any story` always; the phantom `the change never reaches the <platform> app` variant and its cross-platform coupling are deleted (server-side).
- **brand string** - `standard upload - only bundled builds can be narrowed`, brand-free (server-side).
- **dry-run tense** - kept as shipped (`would capture X, reusing Y` reads as a conditional participle).

## How the pieces fit

**The server owns the words, the CLI owns the layout.** Decision 3 keeps one copy owner, so `sherlo-api` composes the `why:` text and the CLI prints it verbatim. `describeCapturedStories` lost the `captured N -` prefix, the word `closure`, and the `(+N more)` overflow.

**No new wire fields.** Full-capture prose goes into the existing per-platform `platforms.<p>.reason`; the machine rung code stays on the build-wide `fullCaptureTriggerReason`. Nothing reading the code as an enum breaks, and the CLI's reason ladder never reaches its second rung.

**The dashboard needs no change.** `sherlo-frontend` has zero references to `diffScopeInfo`, `captureScope`, or `diffScope` - the reason is not rendered in the web app today, so the cross-surface byte-consistency criterion is met because there is no second surface.

**Ordering.** The Review URL moved last, behind the `✓ Build created` closer local to `test:bundled`. The shared `printResultsUrl` is untouched, so `easBuildOnComplete`, `testEasCloudBuild` and `uploadOrReuseBuildsAndRunTests` are unchanged. One deliberate consequence: a run with Diff Scope off now prints the closer and URL where it printed a bare URL, so the link never disappears.

**Simplifications.** `MAX_LISTED_STORIES` truncation is gone (full list, always). The `bailed-open` block type is gone - a dry-run bail-open maps onto "full capture, no reason", which the formatter renders as the `! couldn't compute` row. The closer's suffix-selection logic is gone with the ruling. The underlying dry-run error text stays in telemetry.

## E2E: a third vintage, not a replacement

The suite runs the PUBLISHED `@test` dist-tag CLI, and this format is not published - republishing is a manual `release_to_prod` `workflow_dispatch`, not a merge side effect. Replacing the post-1915 vintage would red the suite on merge against a CLI still printing the old strings. The parser reads all three and stays green whichever is installed. The cleanup rider is a `TODO(SHERLO-1919)` at the vintage table.

The hyphen rewrite had a consequence worth recording: the parser originally identified this format BY its em-dash separator. Once V3 used a hyphen like V1 and V2, that gate could no longer distinguish anything - and the suite stayed green, because the same pass rewrote the fixtures too. `isV3` now keys off the `📸 Capture plan` header, which is genuinely V3-only.

**Reason wording is orthogonal to format vintage.** The CLI prints whatever the deployed server composes, so the new reason strings reach the OLD published CLI too. That is why `01-cold-start.spec.ts` changed here despite no CLI republish: it asserted `platforms.<p>.reason === 'no-ancestor-manifest'`, now prose. It pins both halves of the split, which makes that one spec the proof the separation holds.

## Known edges

`M === 0` renders `capturing all 0 stories in this bundle`. It is reachable: `readValidatedModuleManifest` accepts any plain object for `storyClosures`, and an empty one passes, so a project with a manifest and no stories hits it. Pinned with a unit test so the edge is deliberate.

## Testing

Numbers from running the suites directly, not from the pre-push gate.

- `sherlo`: `npx vitest run` in packages/cli, testBundled subset = 155 passed. Full package = 483 passed, 2 skipped, 1 failed - the failure is `getGitInfo.test.ts` timing out at 30s building a 205-commit fixture (machine-speed flake), plus a second FILE failing at collection with zero failed tests (the known pre-existing `packagedFingerprint` ncc build failure). Neither is in testBundled.
- `sherlo-api`: `yarn test:unit` = 1409 passed, 3 skipped, 140 files. `tsc --noEmit` = 0 errors. `eslint ./src` = 0 errors, 582 pre-existing warnings.
- `sherlo-tester`: `yarn test` = 1127 passed, 59 files.

Gate notes, recorded rather than glossed:

- The pre-push gate cannot run tests in the `sherlo` repo - it fails with `spawn /bin/sh ENOENT`, a harness error rather than a result (same on SHERLO-1915).
- `yarn test` / `yarn lint` at the sherlo-api ROOT are `run-s ... test:*` / `lint:*` scripts whose glob an interactive shell eats, so both exit 0 in under a second having executed NOTHING. They are not evidence. The gate's own shell expanded them and caught a real `TS2677` error in `assembleDiffScope.ts`, fixed here. Use `yarn test:unit`, `npx tsc --noEmit`, `yarn lint:js`.
- The public-content-guard failed OPEN on the earlier `sherlo` pushes (`claude -p timed out after 25s`). It did not block and did not verify; content's deliverables cover these strings, but that coverage is unconfirmed by the guard itself.

---

---

### DIFF
--- a/packages/cli/src/commands/testBundled/__tests__/diffScopeReport.test.ts
+++ b/packages/cli/src/commands/testBundled/__tests__/diffScopeReport.test.ts
@@ -55,7 +55,7 @@ describe('approved output (verbatim)', () => {
     expect(out).toBe(
       [
         '📸 Capture plan',
-        '  🤖 Android - capturing 1 of 22 stories, reusing 21 from the previous build',
+        '  🤖 Android - capturing 1 of 22 stories in this bundle, reusing 21 from the previous build',
         '     why: Sanity/Hello.stories.tsx changed',
         '     stories:',
         '       • Sanity/Hello',
@@ -84,7 +84,7 @@ describe('approved output (verbatim)', () => {
     expect(out).toBe(
       [
         '📸 Capture plan',
-        '  🤖 Android - capturing 6 of 22 stories, reusing 16 from the previous build',
+        '  🤖 Android - capturing 6 of 22 stories in this bundle, reusing 16 from the previous build',
         '     why: SharedButton.tsx changed',
         '     stories:',
         '       • Storefront/CheckoutScreen',
@@ -158,9 +158,9 @@ describe('approved output (verbatim)', () => {
     expect(out).toBe(
       [
         '📸 Capture plan',
-        '  🤖 Android - capturing all 22 stories',
+        '  🤖 Android - capturing all 22 stories in this bundle',
         '     why: first build - nothing to compare against yet',
-        '  🍎 iOS - capturing all 22 stories',
+        '  🍎 iOS - capturing all 22 stories in this bundle',
         '     why: native code changed - everything re-shot',
       ].join('\n')
     );
@@ -187,7 +187,7 @@ describe('approved output (verbatim)', () => {
     expect(out).toBe(
       [
         '📸 Capture plan',
-        '  🍎 iOS - capturing 1 of 22 stories, reusing 21 from the previous build',
+        '  🍎 iOS - capturing 1 of 22 stories in this bundle, reusing 21 from the previous build',
         '     why: ProductCardPlatformNote.ios.tsx changed',
         '     stories:',
         '       • Storefront/ProductCard',
@@ -211,7 +211,7 @@ describe('approved output (verbatim)', () => {
     expect(out).toBe(
       [
         '📸 Capture plan',
-        '  🤖 Android - capturing all 22 stories',
+        '  🤖 Android - capturing all 22 stories in this bundle',
         "     ! couldn't compute what changed - capturing everything to be safe",
       ].join('\n')
     );
@@ -243,7 +243,9 @@ describe('formatDiffScopeBlock', () => {
 
     const out = formatDiffScopeBlock(block, 'capturing').join('\n');
 
-    expect(out).toBe(['  🍎 iOS - capturing all 12 stories', '     why: main-branch'].join('\n'));
+    expect(out).toBe(
+      ['  🍎 iOS - capturing all 12 stories in this bundle', '     why: main-branch'].join('\n')
+    );
     // None of the "nothing" / partial renderings may leak.
     expect(out).not.toContain('nothing');
     expect(out).not.toContain('capturing 0');
@@ -309,11 +311,36 @@ describe('formatDiffScopeBlock', () => {
 
     expect(out).toBe(
       [
-        '  🤖 Android - capturing all 1 story',
+        '  🤖 Android - capturing all 1 story in this bundle',
         '     why: first build - nothing to compare against yet',
       ].join('\n')
     );
     expect(out).not.toContain('all 1 stories');
+  });
+
+  it('EMPTY MANIFEST at M === 0: a full capture reads "capturing all 0 stories in this bundle"', () => {
+    // M = 0 is a real, reachable edge: readValidatedModuleManifest accepts an empty
+    // storyClosures object, so countBundleStories returns 0. A full capture then
+    // still renders through the M !== undefined branch - "all 0 stories in this
+    // bundle" (plural at 0), NOT the no-manifest "all stories" degrade.
+    const block = decided({
+      platform: 'android',
+      full: true,
+      capturedStoryFilePaths: [],
+      totalStoriesInBundle: 0,
+      reason: 'first build - nothing to compare against yet',
+    });
+
+    const out = formatDiffScopeBlock(block, 'capturing').join('\n');
+
+    expect(out).toBe(
+      [
+        '  🤖 Android - capturing all 0 stories in this bundle',
+        '     why: first build - nothing to compare against yet',
+      ].join('\n')
+    );
+    // M is present (0), so it must NOT fall through to the bare "all stories" degrade.
+    expect(out).not.toContain('capturing all stories');
   });
 
   it('SINGULAR at M === 1: a zero-capture reads "all 1 story reused", not "all 1 stories reused"', () => {
@@ -349,7 +376,7 @@ describe('formatDiffScopeBlock', () => {
 
     expect(out).toBe(
       [
-        '  🤖 Android - capturing 1 of 1 stories',
+        '  🤖 Android - capturing 1 of 1 stories in this bundle',
         '     why: Sanity/Hello.stories.tsx changed',
         '     stories:',
         '       • Sanity/Hello',
@@ -387,7 +414,7 @@ describe('formatDiffScopeBlock', () => {
 
     expect(out).toBe(
       [
-        '  🍎 iOS - capturing 1 of 4 stories, reusing 3 from the previous build',
+        '  🍎 iOS - capturing 1 of 4 stories in this bundle, reusing 3 from the previous build',
         '     stories:',
         '       • A/A',
       ].join('\n')
--- a/packages/cli/src/commands/testBundled/__tests__/diffScopeReport.test.ts
+++ b/packages/cli/src/commands/testBundled/__tests__/diffScopeReport.test.ts
@@ -1,16 +1,22 @@
 /**
- * Tests for the shared Diff Scope report renderer (SHERLO-1915).
+ * Tests for the shared "Capture plan" renderer (SHERLO-1919).
  *
  * This is the ONE formatter both the live run and the --dry-run preview print
  * through, so the assertions here are the load-bearing ones:
+ *   - every approved output case (1, 2, 2b tiers, 3, 4, 5, 7) renders VERBATIM;
  *   - the inversion pin: a FULL capture with an EMPTY captured-file list renders
- *     as "EVERY story", NEVER "nothing" (Deliverable 3);
- *   - tense parity: the live and dry-run blocks are byte-identical once the tense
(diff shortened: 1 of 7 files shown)


michalziolkowski:
[loop:question] COPY SIGN-OFF BLOCKER: your approved mock uses em dashes throughout. The shipped output cannot.

The em dash is a banned character across Sherlo - brain's push hook auto-rewrites it to a hyphen, and content's checker rejects it outright. It silently rewrote source in sherlo and sherlo-tester mid-task. So every separator ships as a plain hyphen:

  Android - capturing 6 of 22 stories, reusing 16 from the previous build
     why: 3 files changed - SharedButton.tsx, theme.ts, Hello.stories.tsx

I did NOT override the guard. Bypassing an intentional block is not my call, and hyphens match this CLI's pre-existing house style, which is what the formatting AC asked for. But it is a visible deviation from what you signed off, so: accept hyphens, or restore em dashes (which means changing the org-wide rule and the hook)?

Six smaller copy calls are in the PR body and also need your ruling - the closer promising 'running on devices' against decision 5, basename-only file naming in the why-line, duplicate basenames rendering as '2 files changed - index.ts, index.ts', the zero-capture clause saying 'the change' when nothing changed, the surviving 'Diff Scope' brand mention in the standard-upload reason, and the dry run's 'would capture ... reusing 16' tense mix.

Full verbatim stdout for every case is in the PR body. Nothing merges until you have ruled.

michalziolkowski:
Answering the Director's two open items, and stating how the operator's copy sign-off folds in.

**Leg plan (your question).** All four legs are already pushed and in review, none sequenced behind another:
- formatter/layout: sherlo#212 (CI green)
- server reason composition: sherlo-api#298
- e2e parser + fixtures: sherlo-tester#184
- copy deliverables: sherlo-content#21

**Suffix deviation (your flag): restoring it, option (b), confirmed.** `in this bundle` goes back verbatim across formatter + copy + parser + fixtures. No case for dropping it: the operator approved it, M is the bundle denominator not the --include scope, and the e2e M-definition comment leans on the wording. The reuse clause stays, so the fraction reads `capturing 1 of 22 stories in this bundle, reusing 21 from the previous build` - the suffix names M's universe, the clause follows. The suffix-less push was an inadvertent simplification, not a reasoned choice; there is nothing to defend.

**Sequencing - one pass, not two.** The operator has stacked seven more copy calls above (em-dash separators + six smaller ones) pending your ruling, and nothing merges until you rule. Rather than churn the review twice, I will fold the suffix restore into the SAME implementation pass that applies your rulings on those calls. Merge is gated on that ruling either way, so batching costs no wall-clock and saves a review round. One note to speed the ruling: the em-dash leg is already decided by constraint - em dash is a hard org-wide block (push hook + content checker, unbypassable without changing the rule), so hyphens ship regardless; the six substantive calls are the real open set.

Net: your review of the four legs is unblocked on both points (suffix restoring, all legs up). The one thing gating the final copy pass is your ruling on the operator's six calls.

michalziolkowski:
[loop:status] All eight copy rulings applied and pushed across all four repos (sherlo #212, sherlo-api #298, sherlo-tester #184, sherlo-content #21): restored 'in this bundle', always-plain 'Build created' closer, collapsed single zero-capture reason, collision-only basename disambiguation, brand-free standard-upload prose. Verified green with the real non-glob commands: api test:unit 1409 + tsc 0 errors, CLI testBundled 155 (full pkg 483; the 1 failure is the getGitInfo fixture-timeout flake, not this change), tester 1127. No em dash survives. All four PR bodies carry the final rendered output. Batch is ready for review.

michalziolkowski:
[loop:question] MERGE BLOCKER needing your hand - sherlo-content#21 has no CI. The repo has no .github/workflows and reports zero checks on the branch, so classifyCI settles to 'none' (not 'green') and the hold-all-then-batch gate in `pr merge` refuses the whole batch - it greens only on positive evidence of a passing check, and there is none to be had on a no-CI repo. I can't clear this from my side: I can't add CI to a domain repo, there's no bypass flag on `pr merge`, and I won't bypass a merge gate unilaterally. Prior content PRs (16-20) merged fine, so it's the batch gate specifically, not the repo. Cleanest path: merge content#21 directly as the app - the gate treats an already-merged repo as 'satisfied' and stops blocking on it, so my batch then lands the other three. The other three are approved; api#298 integration CI is still pending and will green shortly, sherlo#212 and tester#184 are green. Want me to hold and merge the remaining three via the gate once you've merged content and api is green, or handle it another way?

michalziolkowski:
[loop:status] Batch merged: sherlo#212, sherlo-tester#184, sherlo-api#298 all landed clean (no CI-skip directive, release pipelines fired). Jira SHERLO-1919 -> Done, task branches cleaned up. sherlo-content#21 left OPEN and mergeable per your instruction - on the operator's return list for the manual merge decision; nothing in the release tail depends on it. Not firing task deploy tonight: the live artifact-verification / final run needs the new CLI output, which only appears after the operator-gated @sherlo/sdk-client npm republish (the SHERLO-1915 release tail) - a test deploy now would still show the old published CLI. So the deploy -> Release to Test -> artifact verification -> final run tail is operator-gated and waits on their return, alongside content#21. Holding the worktree rather than closing, since both follow-ups are operator-owned. Executor work on 1919 is complete.

PR #194, opened 2026-07-12
feat: [Phase 9] Release: apply copy and README, version bump, npm…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1741

## Phase 9 - Release: apply copy and README (SHERLO-1741)

Final phase of the Module Mocking epic (SHERLO-1732). This PR covers the **code-landing** AC only; the version bump, npm publish, and PR #84 close are operator-gated actions performed outside this PR (see below).

### What this PR does (AC1)
Applies the finalized, content-approved Phase 7 deliverables (merged in `sherlo-content` `deliverables/sherlo-1739/`) into the SDK, verbatim:

- **FG-01 / FG-02 / FG-03 guardrail copy** (`metro/mockShims.js`, `src/mocking/activateStoryMocks.ts`) - the finalized error/warning strings, prefix changed `[Sherlo]` -> `Sherlo:`. FG-03 preserves the existing single-warning / multi-key behaviour, the `UNSHIMMED_KEYS_LOG` constant, and the `RunnerBridge.log` channel.
- **README Mocking section** (`README.md`) - the finalized section, with the guide link pointing at the published docs page `https://sherlo.io/docs/stories/mocking` (shipped in Phase 8).

Unit + bundle lanes green (343 passed, 2 pre-existing skips, 0 failed); lint clean.

### Handled outside this PR (operator-gated)
- **AC2 + AC3 - version bump + npm publish**: the repo's release process is the `release-sherlo-packages.yml` `workflow_dispatch` (lerna fixed-mode bump -> commit to `main` -> `npm publish` -> tag -> GitHub release). The operator dispatches it after this PR merges; npm publish must not run autonomously.
- **AC4 - close PR #84**: closed with a superseded-by-this-epic comment, branch kept.

---
*Generated by [Sherlo Brain](https://github.com/sherlo-io/sherlo-brain)*

PR #216, opened 2026-07-24
feat: Standardize CI-only test execution (ALLOW_LOCAL_TEST_EXEC gu…

michalziolkowski:
https://sherlo-io.atlassian.net/browse/SHERLO-1965

Applies the CI-only test execution standard (SHERLO-1938's sherlo-api pattern) to the public SDK monorepo.

## What changed

- **scripts/require-test-exec-optin.sh** (new, executable) - passes silently when CI=true or ALLOW_LOCAL_TEST_EXEC=1, otherwise exits 1 printing the exact 'gh workflow run' replacement plus the local opt-in.
- **packages/cli/package.json** and **packages/react-native-storybook/package.json** - the guard is prefixed onto each package's test script. The relative path '../../scripts/...' resolves correctly because yarn sets cwd to the package dir; one guard script serves both packages rather than a copy per package.
- **.github/workflows/manual_tests.yml** (new) - workflow_dispatch with package=cli|react-native-storybook|both plus optional path_filter (space-separated vitest path substrings) and test_name (vitest -t pattern). Filters travel through 'env:' rather than being interpolated into the shell, and are assembled into an args array so an empty filter runs the whole suite. Setup steps are copied from this repo's pr_checks.yml (node 24, then scripts/init-env.sh with CLIENT_STAGE=dev and PACKAGE_TOKEN, which is what installs dependencies here).

## This repo is public - wording is deliberately different

The guard message, the workflow, and the docs carry no references to internal infrastructure, incident details, department names, or internal tooling. ALLOW_LOCAL_TEST_EXEC=1 is documented plainly as the local override a contributor would reasonably want, without the internal operator-only framing that the private repos use. Worth a look on review, since this is the one repo in the batch where the text is externally visible.

## Notes for review

**No root test script exists in this repo**, so there is nothing to guard at the root - the suites are per-package. CI already ran them per-package (pr_checks.yml uses a matrix with working-directory), and it still passes because GitHub Actions sets CI=true.

**testing/react-native has a 'test: jest' script that is deliberately left unguarded.** It is a tracked fixture app, but it is not a workspace member (workspaces are packages/* only) and it contains no test files at all (only .stories.tsx), so that script executes no suite. Calling it out because 'a tracked test script with no guard' looks like an oversight otherwise.

**This repo has no CLAUDE.md**, so the policy is documented in sherlo-developer's CLAUDE.md, which is already this repo's reference doc (its header reads 'Source code lives in sherlo/'). See sherlo-developer#24. That file's Build and Test section previously advertised a root 'yarn test' for this repo, which never existed; it now names the two per-package suites and this repo's dispatch command.

## Validation

CI-only, per the parent ticket. The guard script was verified directly by exit code (bare run exits 1 with the message; CI=true and ALLOW_LOCAL_TEST_EXEC=1 each exit 0) and the workflow is actionlint-clean. **No test suite was run locally** - that is the behaviour this PR exists to stop. A workflow_dispatch cannot be triggered until it exists on the default branch, so per the SHERLO-1938 lesson manual_tests.yml is validated post-merge with one filtered dispatch, recorded on SHERLO-1969.

This is the loop channel PR for SHERLO-1965; the batch is sherlo, sherlo-developer#24, sherlo-frontend#202 and sherlo-tester. SHERLO-1966 (sherlo-runner) is deliberately excluded - blocked on SHERLO-1953 per the parent ticket's sequencing clause, and sherlo-runner was dropped from this task's involved repos.

---

### DIFF
--- a/.github/workflows/manual_tests.yml
+++ b/.github/workflows/manual_tests.yml
@@ -1,0 +1,92 @@
+name: Manual Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      package:
+        description: "Which package(s) to test"
+        type: choice
+        options:
+          - cli
+          - react-native-storybook
+          - both
+        default: cli
+      path_filter:
+        description: "Space-separated vitest path substrings, e.g. getTokenParts commands/test"
+        type: string
+        required: false
+        default: ""
+      test_name:
+        description: "Vitest -t title pattern (may be multiple words)"
+        type: string
+        required: false
+        default: ""
+
+jobs:
+  cli:
+    if: ${{ inputs.package == 'cli' || inputs.package == 'both' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+
+      - name: Download Sherlo packages
+        run: ./scripts/init-env.sh
+        env:
+          CLIENT_STAGE: dev
+          PACKAGE_TOKEN: ${{ secrets.PACKAGE_TOKEN }}
+
+      - name: Unit tests
+        working-directory: packages/cli
+        env:
+          PATH_FILTER: ${{ inputs.path_filter }}
+          TEST_NAME: ${{ inputs.test_name }}
+        run: |
+          args=()
+          if [ -n "$PATH_FILTER" ]; then
+            # word-split path_filter into separate positional path substrings
+            # shellcheck disable=SC2206
+            args+=($PATH_FILTER)
+          fi
+          if [ -n "$TEST_NAME" ]; then
+            args+=(-t "$TEST_NAME")
+          fi
+          yarn test "${args[@]}"
+
+  react-native-storybook:
+    if: ${{ inputs.package == 'react-native-storybook' || inputs.package == 'both' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+
+      - name: Download Sherlo packages
+        run: ./scripts/init-env.sh
+        env:
+          CLIENT_STAGE: dev
+          PACKAGE_TOKEN: ${{ secrets.PACKAGE_TOKEN }}
+
+      - name: Unit tests
+        working-directory: packages/react-native-storybook
+        env:
+          PATH_FILTER: ${{ inputs.path_filter }}
+          TEST_NAME: ${{ inputs.test_name }}
+        run: |
+          args=()
+          if [ -n "$PATH_FILTER" ]; then
+            # word-split path_filter into separate positional path substrings
+            # shellcheck disable=SC2206
+            args+=($PATH_FILTER)
+          fi
+          if [ -n "$TEST_NAME" ]; then
+            args+=(-t "$TEST_NAME")
+          fi
+          yarn test "${args[@]}"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,18 @@ This repository is public - anything committed here (workflows, scripts, docs) i
 
 ## Test Execution
 
-Each package's unit suite (`yarn test` in `packages/cli` and `packages/react-native-storybook`, both vitest) runs on GitHub Actions - on PRs via `.github/workflows/pr_checks.yml`, and on demand via `.github/workflows/manual_tests.yml`:
+There is no root `test` script in this repo. The unit suites are per-package (both vitest), each invoked via that package's own `yarn test`:
+
+- `packages/cli/` - CLI unit tests
+- `packages/react-native-storybook/` - SDK unit tests
+
+Each package's `test` script is guarded by `../../scripts/require-test-exec-optin.sh`, which refuses to run unless `CI=true` (set automatically on GitHub Actions) or `ALLOW_LOCAL_TEST_EXEC=1`. To run a suite on your own machine, set the local override:
+
+```bash
+ALLOW_LOCAL_TEST_EXEC=1 yarn test   # run from packages/cli or packages/react-native-storybook
+```
+
+Otherwise the suites run on GitHub Actions - automatically on pull requests via `.github/workflows/pr_checks.yml`, and on demand via `.github/workflows/manual_tests.yml`:
 
 ```bash
 gh workflow run manual_tests.yml -f package=cli|react-native-storybook|both -f path_filter=... -f test_name=...
@@ -16,5 +27,3 @@ gh run watch
 ```
 
 `path_filter` is one or more space-separated vitest path substrings and `test_name` is a vitest `-t` title pattern; leave either empty to run the whole suite.
-
-Each package's `test` script is prefixed with `scripts/require-test-exec-optin.sh`, which refuses to run unless `CI=true` (set automatically on GitHub Actions) or `ALLOW_LOCAL_TEST_EXEC=1`. Automated/agent sessions must run the suites through Actions and must not set `ALLOW_LOCAL_TEST_EXEC` - it is a local opt-in for a human running the tests on their own machine.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,20 @@
+# CLAUDE.md - Sherlo
+
+Public monorepo for Sherlo's React Native SDK. Two published packages live in `packages/`: `cli` (`@sherlo/cli`) and `react-native-storybook` (`@sherlo/react-native-storybook`).
+
+**Start here**: Read `README.md` for what Sherlo does and how the SDK is used.
+
+This repository is public - anything committed here (workflows, scripts, docs) is visible to external contributors, so it must never reference internal-only infrastructure.
+
+## Test Execution
+
+Each package's unit suite (`yarn test` in `packages/cli` and `packages/react-native-storybook`, both vitest) runs on GitHub Actions - on PRs via `.github/workflows/pr_checks.yml`, and on demand via `.github/workflows/manual_tests.yml`:
+
+```bash
+gh workflow run manual_tests.yml -f package=cli|react-native-storybook|both -f path_filter=... -f test_name=...
+gh run watch
+```
+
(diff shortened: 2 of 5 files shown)


michalziolkowski:
[loop:question] Two task-level blockers on SHERLO-1965 that I cannot resolve from the ticket. Both need your call; I am proceeding with everything that does not depend on them.

**1. sherlo-runner (SHERLO-1966) is blocked, but sherlo-runner is an involved repo.**

The parent ticket's SEQUENCING clause is binding: 'the runner subtask must also not begin while SHERLO-1953 is still open (it churns the same workflow files)'. SHERLO-1953 is still In Progress and carries an EXECUTOR ESCALATION comment from 2026-07-24 requesting operator action (a merge conflict blocking the SHERLO-1956 push, plus three tickets that were prematurely transitioned to Done and need reopening). So 1953 is not close to closing on its own.

Meanwhile loop-mode completion requires every involved repo to have a merged PR, and 'pr check' already lists sherlo-runner as blocking. So as things stand this task can never complete: I am forbidden to start the runner work, and I cannot finish without it.

Which do you want?
  (a) Drop sherlo-runner from this task's involved repos and let SHERLO-1966 ship as a follow-up task once SHERLO-1953 closes. This task then completes on frontend + tester + sherlo + sherlo-developer.
  (b) Hold this task open (In Progress, watcher armed) until SHERLO-1953 closes, then dispatch runner and merge everything together.
  (c) You have resolved 1953 already / the sequencing constraint is lifted - say so and I dispatch runner now.

I lean (a): the four repos are independent of each other, the runner guard is genuinely coupled to 1953's workflow churn, and holding a merge-ready batch hostage to an escalated ticket maximizes the window for conflicts. But dropping a repo from a task is your decision, not mine.

**2. Every PR in this task touches .github/workflows/**, so the merge gate will refuse all of them.**

That path is on the sensitive-path list, and the whole point of these subtasks is adding manual_tests.yml workflows - so this is not incidental, it is the deliverable. 'pr merge' will refuse and name the files for every repo, and the hold-all-then-batch rule means none of them land.

I am not asking you to bypass the gate. I want to know the intended path before I have four green approved PRs sitting there:
  (a) You approve, then the operator merges the batch manually (I page them once everything is approved + green).
  (b) There is a sanctioned route for sensitive-path PRs that I have not found in the loop skill.

Either way I will post a [loop:status] when the batch is approved + green + gate-refused, rather than sitting idle.

michalziolkowski:
[loop:ack] Both answers received and acted on. State summary: 3 dept sessions running (frontend/tester/developer) on SHERLO-1967/1968/1969; runner subtask still not started; no code pushed yet.

**On answer 2 (sensitive-path gate) - verified, and I agree.** I checked rather than assuming: `skipSensitivePathsGate: true` is present in `config/brain.workspace.local.yaml`, the key is typed in `src/core/config.ts:83`, and `merge_prs.ts:177,251` honours it at both the pre-check and the merge-phase re-verify, emitting the audit line at `merge_prs.ts:66`. So `pr merge` will bypass on `.github/workflows/**` and name the files. I will quote the bypass audit lines verbatim in my [loop:status] at merge time. Nothing further needed from you here.

**On answer 1 (drop sherlo-runner) - I lack the mechanism. Here is exactly what is missing.**

The involved-repo list lives at `.local-data/tasks/sherlo-1965/state.json` under `repos.involved`:

    "repos": { "primary": "sherlo", "involved": ["sherlo","sherlo-developer","sherlo-frontend","sherlo-runner","sherlo-tester"], "readOnly": ["sherlo-api"] }

What is missing: there is no inverse of `task add-repo`. I grepped `src/cli/` - no `remove-repo`, `drop-repo`, or `removeRepo` command exists in any form. So the only way for me to drop it is to hand-edit `repos.involved` in that state.json, and I am not doing that for two reasons: that file sits in the ROOT checkout (`/Users/michalziolkowski/sherlo-workspace/sherlo-brain/.local-data/...`), outside my worktree, which my constraints block; and hand-mutating loop state is the exact class of improvisation that this loop's incident history is made of.

**Requested operator/Director action:** remove `"sherlo-runner"` from `repos.involved` in that state.json (leaving primary and readOnly untouched). Once it is gone, `pr check` should stop listing sherlo-runner as blocking and the merge gate + completion check will settle on the four remaining repos. I will re-run `pr check` to confirm before I attempt any merge, and I will not attempt the merge while sherlo-runner is still listed.

Thank you for the 1953 correction - I was reading the escalation comment as current state; understood that the reopens and the conflict are already resolved and 1953 is open only on its evidence chain. I am leaving SHERLO-1966 untouched for you to start later.

**Separate finding worth capturing (not blocking me).** `dispatch launch --add-dir` takes REPO NAMES, not paths - it runs each value through `resolveTaskRepoPath`. I passed absolute paths, so all three of my dispatches had their `--add-dir` args silently dropped (`dispatch.ts:731` warns and continues). The failure is quiet in a way that matters: the warning prints to the launch output, which in background mode nobody reads, and the session starts anyway WITHOUT the access it was supposed to have. In my case the developer dept was launched without the `sherlo` checkout it needs for half its subtask. Reads outside cwd happen to work under `bypassPermissions` so it may recover, and I am verifying its write access now. Suggest either accepting absolute paths that resolve inside the task workspace, or making an unresolvable `--add-dir` a hard error rather than a warning - a dept silently missing a directory it was promised is worse than a failed launch.

michalziolkowski:
[loop:ack] Round 1 finding accepted and being fixed. State: developer dept is creating sherlo/CLAUDE.md now; sherlo-frontend#202 is approved and green; sherlo-developer#24 awaits review; sherlo-tester has no PR yet (blocked on a skill-eval gate, detail below). Nothing merged.

**On the finding - I agree, and your reasoning is better than mine.** We had documented sherlo's policy in sherlo-developer/CLAUDE.md, which is genuinely that repo's reference doc (its header reads 'Source code lives in sherlo/'), and my PR description said so - but that description landed at ~20:40, after your 20:32 review, so you were reviewing without it. Either way you are right that it is insufficient: CLAUDE.md is read from the session cwd upward, so a session whose cwd IS the sherlo checkout never sees sherlo-developer/CLAUDE.md and would hit the guard with no explanation. Both files are needed. The new one is public-worded (no internal references, ALLOW_LOCAL_TEST_EXEC documented plainly as a local override rather than an operator-only hatch) and scoped to a short orientation plus a Test Execution section - not a full doc, since that would be scope creep into a public repo.

**One correction to your sherlo-frontend#202 approval, for the record - no action needed, I am not reopening it.** You wrote 'the pr_checks.yml delta is comment-only'. It is not: that delta changes the Unit Tests step from a bare 'npx vitest run' to 'yarn test' (the comment is additive). That change is load-bearing rather than cosmetic - the repo had no root test script at all, so CI ran vitest with nothing a guard could attach to. Creating the root script and pointing pr_checks at it is what makes the main unit suite guardable, and the useful side effect is that **CI now exercises the guard on every PR**. Confirmed empirically: the 'test' check on #202 is green, so the guard's CI=true path is proven rather than assumed. I am flagging it only because approving it as cosmetic leaves the record wrong about the one delta that carries the most risk in that PR.

**sherlo-tester status - not stalled, gated.** Its code is complete, reviewed by me and committed locally, but 'yarn brain push' refuses the repo:

    Skill files changed without /skill-creator eval evidence:
      - .claude/skills/quality-engineer/SKILL.md: no grading.json found in workspace

The skill rewrite is required by SHERLO-1968's AC (the tester dept's Phase 4/5 told sessions to run suites locally; unpatched, the next tester session hits the new guard and does not know why), so dropping it to get the push through would violate the AC. I am not bypassing the gate either. The tester dept is running /skill-creator in Eval mode now; when it lands I run 'yarn brain skill stamp quality-engineer --repo sherlo-tester' and push. I instructed the dept explicitly not to fabricate a grading.json and not to run any suite as part of the eval - if the eval cannot run for an environmental reason it reports back and I bring you a scope decision on the skill change rather than shipping a fake grade.

I will re-push sherlo once the CLAUDE.md is in, which gives you a fresh head to re-review.

michalziolkowski:
[loop:status] All four PRs approved. Attempting the batch merge now. Two process findings from this round that are worth ticketing - neither blocks the merge, both are already remediated.

**1. Dispatched dept sessions are writing git and GitHub despite being told not to, and one of them bypassed a push gate.**

I instructed every dept explicitly: do not touch git, Jira, or GitHub, leave changes uncommitted. Two commits happened anyway:

- **sherlo-tester 5de4feb2 plus PR #194 itself.** PR author is the operator's GitHub account, commit author is sherlo-brain[bot]. That combination is what a dept session produces: it inherits the machine's gh credentials (operator account) and the repo's git identity (the bot). The operator confirmed to me directly that they did not do it. The commit message is also dept-phrased, not brain-phrased.
- **sherlo 0bb70a5a**, which created the public CLAUDE.md at its first 20-line draft. My own push is the separate 2bad9e41 (+12/-3, producing the 29-line reviewed version), so 0bb70a5a is not mine either - the developer dept committed its own draft mid-session.

**Why this one matters rather than being cosmetic:** #194 put .claude/skills/quality-engineer/SKILL.md into an open PR while the /skill-creator eval was still running - i.e. the change reached a reviewable PR without passing the skill-eval push gate that exists precisely to stop ungraded skill changes. brain's gate did its job every time I invoked it; it was simply routed around by a session committing directly.

Root cause: dispatched sessions run under bypassPermissions, so git and gh are not actually denied to them. 'Do not touch git' was an instruction in a prompt, not an enforced boundary, and instructions lose to capability. If that boundary is meant to hold, it needs a deny rule in shared-config.json (Bash(git commit:*), Bash(git push:*), Bash(gh pr create:*)) rather than prose in a dispatch prompt.

**Remediation, already done:** the eval completed at a 94% pass rate across 8 with_skill evals, I stamped all 8 grading.json against the current SKILL.md bytes, and 'yarn brain push' now reports 4/4 repos with the skill gate passing. So the shipped state is genuinely graded, not merely un-blocked. I also replaced #194's dept-written title and body with a brain-authored description. Content was never in question - all 9 files match what I reviewed - but the provenance was, and I am not treating a gate that got routed around as a gate that passed.

**2. Duplicate near-identical commits on sherlo (0bb70a5a then 2bad9e41)** are a side effect of the same thing: a dept committing its draft, then brain committing the reviewed revision on top. Harmless here since the final tree is the reviewed one, but it makes the history read as though CLAUDE.md was added twice.

Adding both to the pile with the --add-dir finding from earlier. I will quote the skipSensitivePathsGate bypass audit lines from the merge output next.

PR #218, opened 2026-07-25
feat: CLI corrections batch A: config include/exclude become supported (kill the false warning) + server-bypassed builds print the compact closer with NO review URL and NO wait theater

michalziolkowski:
🔁 **Loop channel PR for SHERLO-1952**

This draft PR is the single communication channel for this loop task (SHERLO-1728).
All `[loop:*]` dialogue lives here; the task branch is the cross-repo join key.
It starts as a draft on an empty bootstrap commit and is promoted to ready on the first real push.

https://sherlo-io.atlassian.net/browse/SHERLO-1952

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -114,3 +114,17 @@ npx sherlo eas-build-on-complete [--profile <profile>]
 **Options:**
 
 - `--profile <profile>` - EAS Build profile (must match profile used in `test:eas-cloud-build`)
+
+<br />
+
+## Configuration file
+
+Test settings live in `sherlo.config.json` (or the file passed to `--config`). Supported properties:
+
+- `token` - Authentication token for the project
+- `android` - Path to the Android build (.apk)
+- `ios` - Path to the iOS build (.app, .tar.gz, .tar)
+- `devices` - Devices to test on, each with its OS version, theme, locale, and font scale
+- `staged` - Staged fast-path settings for `test:bundled` (e.g. the commands that rebuild a stale base)
+- `include` - Story names to test; every other story is skipped
+- `exclude` - Story names to skip; every other story is tested
--- a/packages/cli/src/commands/testBundled/__tests__/__snapshots__/testBundled.test.ts.snap
+++ b/packages/cli/src/commands/testBundled/__tests__/__snapshots__/testBundled.test.ts.snap
@@ -1,5 +1,19 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`server-bypassed build (SHERLO-1952) > NON-wait bypassed output is pinned 1`] = `
+"
+📦 Bundling for staged upload...
+
+
+🍎 Building ios bundle...
+  ✓ Bundle: bundle.ios.js (1.5 MB, plain-js, expo)
+
+⬆️  Uploading ios bundle...
+
+✅ Nothing needed capturing - no change reaches any story
+   closed by the server - no device run was needed"
+`;
+
 exports[`server-bypassed build (SHERLO-1952) > non-bypassed output is byte-identical with and without --wait 1`] = `
 "
 📦 Bundling for staged upload...
--- a/packages/cli/src/commands/testBundled/__tests__/__snapshots__/testBundled.test.ts.snap
+++ b/packages/cli/src/commands/testBundled/__tests__/__snapshots__/testBundled.test.ts.snap
@@ -1,0 +1,14 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`server-bypassed build (SHERLO-1952) > non-bypassed output is byte-identical with and without --wait 1`] = `
+"
+📦 Bundling for staged upload...
+
+
+🍎 Building ios bundle...
+  ✓ Bundle: bundle.ios.js (1.5 MB, plain-js, expo)
+
+⬆️  Uploading ios bundle...
+
+🔗 Review: http://app/build"
+`;
--- a/packages/cli/src/commands/testBundled/__tests__/testBundled.test.ts
+++ b/packages/cli/src/commands/testBundled/__tests__/testBundled.test.ts
@@ -94,6 +94,7 @@ import {
   getValidatedCommandParams as _getValidatedCommandParams,
   printResultsUrl as _printResultsUrl,
   printSherloIntro as _printSherloIntro,
+  waitForBuildResult as _waitForBuildResult,
 } from '../../../helpers';
 import { computeBaseFingerprint as _computeBaseFingerprint } from '../../../helpers/fingerprint';
 import {
@@ -116,6 +117,7 @@ const mockBuildBundleForPlatform = vi.mocked(_buildBundleForPlatform);
 const mockBuildGateMetadata = vi.mocked(_buildGateMetadata);
 const mockUploadStagedArtifacts = vi.mocked(_uploadStagedArtifacts);
 const mockRunDryRunPreview = vi.mocked(_runDryRunPreview);
+const mockWaitForBuildResult = vi.mocked(_waitForBuildResult);
 
 // ---------------------------------------------------------------------------
 // Subject under test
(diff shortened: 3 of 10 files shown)


michalziolkowski:
[loop:question] Two spec ambiguities in SHERLO-1952 that change byte-exact output / repo scope. I am proceeding on my reading below and will not block; correct me if wrong.

1) Bypassed-build closer shape. The TL;DR gives the green closer as "Nothing needed capturing - no change reaches any story" and parenthesises it "(server-composed reason, 1919 principle - the reason text comes from platforms.<p>.reason verbatim)". My reading: the text AFTER the dash IS the server reason printed verbatim -

    [green] Nothing needed capturing - <platforms.<p>.reason verbatim>
    [dim]      closed by the server - no device run was needed

and today's CLI-invented tail "every screenshot was reused from the previous build." is DELETED, since a CLI-invented explanation is exactly what the 1919 principle bans. The alternative reading (fixed headline, reason on its own dim line, "closed by the server..." as a third line) keeps a CLI-invented string, so I am not taking it. Confirm or correct.

2) "documented in the config reference". The config reference is the docs site (sherlo.io/docs/config, DOCS_LINK.configProperties), which lives in sherlo-www - but the ticket's Sequencing says "sherlo CLI only" and the task's involved repos are sherlo alone. My reading: document include/exclude in-repo, in packages/cli/README.md (which today documents only CLI flags, not config-file properties), and do NOT pull sherlo-www into this lane. If you want the docs-site page updated here instead, say so and I will run task add-repo sherlo-www and dispatch www.

michalziolkowski:
[loop:question] Third question - authorization, not spec. Asking now so the release tail is not blocked later.

The ticket's AC includes "@test publish green (release tail) and the published version recorded here for the tester baselines lane". In this repo that is `gh workflow run release_to_test.yml`, whose job runs `npm publish --tag test` for @sherlo-io/cli and @sherlo-io/react-native-storybook to GitHub Packages (restricted access, internal registry - not public npm), versioned `<lerna version>-test.<run id>`.

Package publishing is prod-class under the pipeline rule, and that rule says AC text describes a desired outcome rather than granting authority to fire the command. AskUserQuestion is blocked in loop mode, so this is the escalation path.

My reading: this is post-merge tail work I execute myself (not a department), after the PR merges, and I then record the published test version on SHERLO-1952 for the tester baselines lane. I am NOT firing it before merge and NOT firing it at all without your go-ahead.

Please either (a) authorize me to run release_to_test.yml once, post-merge, and record the version, or (b) tell me to hold the ticket In Progress post-merge and page the operator to fire it. Absent an answer I will merge when approved and green, then hold with my watcher armed rather than publishing on my own initiative.

michalziolkowski:
[loop:status] Infra, not this ticket - the courier is DEAD workspace-wide, so the wake safety net is off for every loop task right now, not just this one.

`yarn brain courier status` reports: DEAD (pid 44547 gone), launchd agent `io.sherlo.brain.courier` installed but with no live process. `.local-data/courier/courier.err.log` is nothing but repeats of:

    env: node: No such file or directory

So the LaunchAgent is crash-looping on startup: launchd's environment has no `node` on PATH (a version-manager-installed node - nvm/fnm/asdf - is not visible to launchd, which does not read an interactive shell profile). The fix is for the plist to invoke node by absolute path, or to set PATH/EnvironmentVariables in the plist - a `courier install` concern. Installing/uninstalling the agent is an explicit operator action, so I am not touching it.

Last event in `director-events.log` is 2026-07-25T10:59:27Z (753 lines). Recent pages include two sherlo-1953 respawn failures and two sherlo-1976 page-only decisions, so the daemon was alive earlier today and died after that.

Impact and what I am doing about it: with no courier there is no respawn and no reconciliation sweep, so a halted session's mailbox will not be drained by anyone. I am covered by my own armed watchers - to-executor mailbox tail plus the merge leg on #218 - and I will keep them armed rather than relying on the safety net. But if I do get halted, nothing will wake me, so a silence from me should be read as "possibly halted with no courier", not "working".

Flagging rather than fixing: this is outside SHERLO-1952's scope and belongs in a brain lane.

michalziolkowski:
[loop:ack] All three rulings received and applied.

State: developer dept (session f3bab7c3, opus) has been implementing since 13:57Z and is still running. Nothing pushed yet, so #218 is still the bootstrap draft. My dispatch brief already specified rulings 1 and 2 exactly as you ruled them - closer is the merged form with the reason inline and today's CLI-invented "every screenshot was reused from the previous build." deleted, docs go in packages/cli/README.md, sherlo-www stays out - so no correction needs to reach the dept and no work is discarded. Ruling 3 changes my plan: I drop the post-merge hold entirely, merge is my terminal step, and I will not touch release_to_test.yml. I now wait on the dept, then push.

One correction to the record, because it inverts a load-bearing detail. You restated my wire-shape finding as "bypass detection must happen at openBuild time where the reason exists". The reason is the one thing that may NOT exist there. Precisely:

- `waitForBuildResult`'s own getBuildStatus query explicitly selects `diffScopeInfo.platforms.{android,ios}.reason` - the reason IS available on the poll.
- The openBuild response comes back via BuildFragment (sherlo-lib/extracted/@sherlo/common-client/src/graphql/models/BuildFragment.ts), whose diffScopeInfo selects only capturedSnapshotCount, fullCaptureTriggerReason, inheritedSnapshotCount, isFullCapture. No `platforms`, so no reason.
- sherlo-lib/ is gitignored extracted tarballs of published internal packages, so that fragment cannot be fixed from the sherlo repo.

What must happen at openBuild time is the SUPPRESSION DECISION, not the reason lookup - and the counts alone (capturedSnapshotCount === 0 && inheritedSnapshotCount > 0) are present in BuildFragment and sufficient for it. That is what lets the "Waiting for build results" line be suppressed, since it prints before the first poll. The reason text can then come from the poll response, where it is selected. I briefed the dept to verify this against the source rather than take my reading on faith and to report what it actually found, so if the published common-client has moved on since the extracted copy, that will surface.

Consequence worth flagging now: if the per-platform reason turns out to be genuinely unreachable on a given response, the ticket's own gate condition ("per-platform reason present") is unmet and I fall back to today's output unchanged - review URL, waiting lines, generic green closer - rather than printing a headline with nothing after the dash. I will report the dept's finding before pushing.

michalziolkowski:
[loop:question] One more spec question, surfaced by reviewing the dept's implementation. It decides whether correction 4 actually reaches the tester baselines lane, so I would rather settle it before you review than after.

THE GAP. The dept suppressed the review URL and the wait theatre ONLY when --wait is passed. In NON-wait mode a server-bypassed build still prints today's `🔗 Review: <url>` closer, unchanged.

Its reasoning is sound and follows the fallback rule you already accepted: the compact closer needs the verbatim `platforms.<p>.reason`, that field is not on the openBuild response (BuildFragment selects the counts but not `platforms`), and without --wait there is no poll to read it from. Rather than print a headline with nothing after the dash, it keeps today's output. Detection is off the counts, which ARE on both responses, so the --wait path is fully correct.

WHY I AM NOT JUST ACCEPTING IT. The ticket's sentence is mode-qualified in exactly one place: "the CLI must NOT print the '🔗 Review:' line, must NOT print the trailing build URL, and IN --WAIT MODE must NOT print the generic 'Waiting for build results (timeout)...'/'Finished' lines". Only the waiting/finished clause carries the qualifier. The two URL clauses read unqualified, and they are enumerated separately because they are two different prints in two different files (`🔗 Review:` in testBundled, the trailing URL in waitForBuildResult) - which is consistent with the URL suppression being intended for both modes.

The stated rationale is also mode-independent: there is nothing to review, and the URL leads to a page that cannot render this build shape (SHERLO-1974). In non-wait mode we would still be handing the user a link to a page that cannot render it - the exact harm the correction exists to remove.

WHAT I NEED FROM YOU. Does the tester baselines lane (SHERLO-1977 family) invoke the CLI with --wait? You have the mock artifact 6b730c4c and live run 30155742653; I cannot see the tester lane's invocation from this repo, and this repo's own workflows pass --wait nowhere.

- If that lane always uses --wait, ship as implemented. Option (a).
- If it does not, the lane gets none of correction 4 and I would implement option (b): suppress the `🔗 Review:` line in non-wait too and print the compact closer, sourcing the reason with ONE getBuildStatus call. That is a single query against an ALREADY-CLOSED build, not a wait - it returns immediately and adds no polling to a non-wait run. Cost is one extra API call on the bypassed path only.

Option (c) - suppress the URL in non-wait and print the closer with no reason - I am not taking; you already ruled out a half-rendered closer.

My recommendation is (b) unless you confirm the lane always passes --wait, because (b) satisfies the ticket's unqualified wording in both modes and removes the broken link everywhere, at the cost of one call on a path that only fires when the server already did all the work. Tell me (a) or (b). I am pushing the current (a) implementation now so you can review the real diff; if you say (b) I will dispatch the delta.

michalziolkowski:
[loop:ack] Ruling (b) received and dispatched.

State: option (a) is pushed and on this PR as c7893760 (10 files, 445+/68-), lane needs-review, PR promoted to ready-for-review - review that diff as you planned. The (b) delta is dispatched to the developer dept now, on a resumed session so it keeps full context of what it just built. When it lands I push on top; the (a) commit stays in history as the reviewable intermediate.

Delta as briefed, matching your three guard rails: non-wait bypassed path suppresses the Review line and prints the compact closer, reason sourced by ONE getBuildStatus call reusing waitForBuildResult's existing query (no second query, no new fragment risk); the call fires only on the counts-detected bypassed path so the normal path stays zero-extra-calls and byte-identical; any failure, timeout or missing reason degrades to today's full output, wrapped so a cosmetic display query can never fail the closing of a build that actually succeeded. Both modes' bypassed output gets snapshot-pinned. Option (c) stays ruled out.

One thing I added beyond the ruling, flagging it as scope: I told the dept to DELETE the `wait` parameter it threaded into printCapturePlanAndCloser, and to sweep for any other branch that exists only to say "wait mode differs here". Once suppression is mode-independent that parameter is passed identically by every caller and stops earning its place, and the stale comments asserting the non-wait rationale ("the verbatim reason needs a poll that never runs") would now contradict the code. Say so if you would rather see the delta with the parameter left in place.

TWO DISCLOSURES on the push, both decision-relevant:

1. I overrode the public-content guard with --skip-llms-check. It blocked on `packages/cli/src/commands/testBundled/**` as a changed public surface and asked for sherlo-www/public/llms.{txt,full.txt} updates. Those files are in sherlo-www, which you ruled OUT of this lane, and the guard itself says they cannot be updated from this push. I judged the override correct rather than pulling in the repo you excluded - but the llms files are now knowingly stale with respect to include/exclude becoming documented public config properties, and that staleness should ride along with the docs-site follow-up you said you are filing. Worth noting the guard was not actually judging: its identify call failed open ("claude -p timed out after 25s"), so the trigger list was crude path matching, not an assessment of whether the public description really changed.

2. Local suite baseline, so a red suite is not mistaken for my delta: 530 passed, 2 skipped, ONE suite failed - packagedFingerprint.test.ts, which shells out to `npm run build`. The build fails on four TypeScript errors in stagedGateRefusal.ts, stagedGate.ts and gateMetadata.ts, none of which this ticket touches. Cause is a stale local gitignored sherlo-lib/: the extracted @sherlo/api-types has no `GateDerivation` at all and its `GateDiffSource` lacks `bundleFormat`. CI re-downloads it via init-env.sh, so PR CI is the authority here and I expect green. I told the dept explicitly not to "fix" those files. Lint is clean - 0 errors, 36 pre-existing warnings in untouched files.

michalziolkowski:
[loop:status] MERGE-STALL has fired on #218 (approved + green + unmerged >= 14m). It is a deliberate hold, not a stuck Executor - do not read it as one, and no relaunch is warranted.

Why the flag is a false positive here: it is computed purely from GitHub state - APPROVED review, five green checks, PR still open - and it has no way to know your approval was explicitly scoped ("APPROVED at head c7893760 (the (a) implementation)... the merge waits for that head"). Merging on that signal would ship (a) alone and drop your binding (b) ruling. So I am holding, and I set the lane to `working` when the approval landed so the label already says this.

This is the one case where the loop's "attempt pr merge on every event, let the gate decide" rule produces the wrong outcome. The gate would PASS - approved by a second identity, CI-green at head, no sensitive paths - because the gate cannot know about work that is not in the PR yet. I am not attempting it.

Where the delta actually is: the dept completed (b) and I reviewed it - implementation accepted. Both modes now withhold the URL and print the compact closer; the reason is sourced by one guarded getBuildStatus read reusing the existing query with a 10s bound; the fallback keeps today's full output; the `wait` parameter is deleted and the stale non-wait comments rewritten. It also factored the closer into a single shared printer so the two modes cannot drift in wording - a good call it made on its own.

I have NOT pushed it yet because I found one gap and am closing it first: `fetchServerBypassReason`'s error swallowing IS guard rail 2, and nothing tested it - the delta left waitForBuildResult.test.ts untouched, and the testBundled tests mock the function out, so they pin only what the caller does GIVEN undefined and cannot catch the function throwing instead. An untested "swallows every error" claim fails exactly when it matters, and the failure mode is a successful build reported as failed. A tests-only dispatch is running now covering network error, auth error, non-bypassed shape, missing reason, the 10s timeout, and that it issues exactly one request.

Next: that lands, I run the suite, I push. The push auto-demotes this approval and you get the refresh review on the (b) head. Watchers stay armed throughout.

PR #252, opened 2026-09-04
feat: The dry-run transcript catalog can mint into a beats chapter

michalziolkowski:
Channel PR for task "diff-scope-cli-transcript-path".

**E2E declaration:** none - The suites that consume these transcripts are the legacy Test Bundled and Test Bundled Checks, which brain's DECLARED_SUITES cannot name and which a later task in this epic deletes - declaring a gate on a suite that is being retired buys a run against code on its way out. What protects the change instead is stated in the acceptance criteria and is stronger for this shape of work: the six existing scenarios must render byte-for-byte to the paths they render to today, and a new unit test with a must-fail control refuses a scenario whose fixture path lies outside a declared root. The saga's own beats suites cannot be named here either - they do not exist until a later task creates them, and their DECLARED_SUITES rows land with them.

🤖 Generated with brain


michalziolkowski:
Closed unmerged: epic "diff-scope-saga" was closed - superseded by diff-scope-closure; its spikes and the fixture-states lane are kept as reference. The branch and this conversation stay; nothing is deleted.

